PR #3, opened 2018-01-09
TX/RX Handler statistics

gippy:


src/handler_base.js
```diff
@@ -170,6 +170,15 @@ export default class HandlerBase extends EventEmitter {
         }
     }
 
+    getStatistics() {
```
jancurn: Maybe return null rather than undefined, so that users know which fields should be there but are missing.

src/server.js
```diff
@@ -357,6 +358,18 @@ export class Server extends EventEmitter {
         .nodeify(callback);
     }
 
+
+    /**
+     * Calls handler by ID and reads it's usage statistics.
+     * @return {Object} statistics { srcTxBytes, srcRxBytes, trgTxBytes, trgRxBytes }
+     */
+    getStatisticsForHandler(handlerId) {
```
jancurn: Let's rename "Statistics" to "Stats", it's easier to write :)

jancurn: And maybe I'd rename this to getConnectionStats(connectionId), it will make it more clear for users.

src/server.js
```diff
@@ -209,6 +208,7 @@ export class Server extends EventEmitter {
                 socket.pause();
 
                 const funcOpts = {
+                    handlerId: result.id,
```
jancurn: handlerId => connectionId - see below

src/handler_base.js
```diff
@@ -170,6 +170,15 @@ export default class HandlerBase extends EventEmitter {
...
 
+    getStats() {
+        return {
+            srcTxBytes: this.srcSocket ? this.srcSocket.bytesWritten : undefined,
+            srcRxBytes: this.srcSocket ? this.srcSocket.bytesRead : undefined,
+            trgTxBytes: this.trgSocket ? this.trgSocket.bytesWritten : undefined,
+            trgRxBytes: this.trgSocket ? this.trgSocket.bytesRead : undefined,
+        };
+    }
+
```
jancurn: please change undefined to null, it's more transparent (e.g. if you call JSON.stringify)

PR #4, opened 2018-01-22
Added close connection event

gippy:


src/server.js
```diff
@@ -253,10 +264,26 @@ export class Server extends EventEmitter {
 
     handlerRun(handler) {
         this.handlers[handler.id] = handler;
+
         handler.once('destroy', () => {
```
jancurn: Is handler for 'destroy' necessary? 'handlerClosed' will be called anyway, right?

src/server.js
```diff
@@ -226,17 +231,23 @@ export class Server extends EventEmitter {
...
+
                 // If not authenticated, request client to authenticate
                 if (funcResult && funcResult.requestAuthentication) {
+                    // connection was closed immediately
+                    this.log(`${result.id} - closed connection`);
+                    this.emit('connectionClosed', {
+                        connectionId: result.id,
+                        stats: { srcTxBytes: 0, srcRxBytes: 0 },
+                    });
                     throw new RequestError('Proxy credentials required.', 407);
```
jancurn: Actually I thought that 'connectionClosed' shouldn't be emitted more than once as it will be confusing for callers. Here, for example, the event will be emitted once and then RequestError will cause failRequest() to emit the event again.

src/server.js
```diff
@@ -274,6 +286,14 @@ export class Server extends EventEmitter {
             this.sendResponse(request.socket, 500, null, 'Internal error in proxy server');
             this.emit('requestFailed', err);
         }
+        // emit connection closed if request fails and connection was already reported
+        if (handlerOptions) {
+            console.log(`${handlerOptions.id} - closed connection`);
```
jancurn: No console.log() please :)

src/server.js
gippy: connectionClosed here is emitted because no handler was actually created (it fails in constructor of handler) but it is already reported to the main proxy component in prepareRequestFunction. Basically this is just a cleanup call. If this wasn't here than any calls that correctly assign auth options but incorrectly set upstreamProxy would fail and the connection will be never marked as closed.

src/server.js
gippy: I cleaned it up a little, but had to keep the destroy event. It would take longer time to figure out why it's needed, but some of the tests were failing when I moved all of the functionality to  handlerClose and removed destroy event. Will take a look at it later.

src/server.js
jancurn: Yeah, no handler is created but `this.failRequest(request, err, handlerOptions);` will be called anyway, in both `onRequest()` and `onConnect()`, and it emits the event again

gippy: Unless I'm mistaken for each connection either onRequest or onCalled is called, not both. And for each function it can call failRequest 3 times:

1) When prepareRequestFunction throws error, then handlerOptions is empty and failRequest does not emit "connectionClosed"
2) When handler is constructed, which will call failRequest with handlerOptions, so it will emit "connectionClosed" 
3) When it fails during runHandler, which results in the same as during constructor. 

Only if both constructor and runHandler are processed, are events for 'close' and 'end'  processed which in turn can call "connectionClosed", so there is no way that both failRequest and Handler.emitHandlerClosed are called.

I added an unit test which checks it this for all available tests and it looks like it works.

jancurn: Oh I see, sorry, didn't notice this before.

PR #16, opened 2018-06-12
Fix: Throw RequestError when request.url cannot be parsed

gippy:


src/server.js
```diff
@@ -236,6 +236,9 @@ export class Server extends EventEmitter {
 
                     this.stats.httpRequestCount++;
                 }
+                if (!handlerOpts.trgParsed) {
```
jancurn: I think the crash can happen earlier above - see:
`const parsed = parseUrl(request.url);`

jancurn: and `if (!parsed.protocol)` below that

gippy: Oh you are right, weird that it does not crash there, from checking logs it always crashes here, but you are right, I will add the condition there too,

PR #31, opened 2018-12-20
Added event 'onTrgRequestConnect' forwarding to get headers from CONNECT event

Benraay:
I have added an event 'onTrgRequestConnect' forwarding to the server
and also added an helper 'listenConnectAnonymizedProxy' inside anonymize_proxy
so we can add a callback on the anonymized proxy


Benraay:
Ok I will do it a bit later thanks for your reply

PR #33, opened 2018-12-26
Fix #32 - tunnel chain should set the target host in the 'host' heaer and not in the path parameter

### LINKED ISSUE #32: tunnel chain should set the target host in the 'host' header and not in the path parameter
The handler_tunnel_chain.js file creates a `CONNECT` request to the real proxy and sets the desired target in the path parmater of the request options.
As far as I know the standard it to set it in the 'host' header...

regevbr:
The handler_tunnel_chain.js file creates a CONNECT request to the real proxy and sets the desired target in the path parmater of the request options.
As far as I know the standard it to set it in the 'host' header...

### DIFF
--- a/src/handler_tunnel_chain.js
+++ b/src/handler_tunnel_chain.js
@@ -20,12 +20,16 @@ export default class HandlerTunnelChain extends HandlerBase {
     run() {
         this.log('Connecting to upstream proxy...');
 
+        const targetHost = `${this.trgParsed.hostname}:${this.trgParsed.port}`;
+
         const options = {
             method: 'CONNECT',
             hostname: this.upstreamProxyUrlParsed.hostname,
             port: this.upstreamProxyUrlParsed.port,
-            path: `${this.trgParsed.hostname}:${this.trgParsed.port}`,
-            headers: {},
+            path: targetHost,
+            headers: {
+                host: targetHost,
+            },
         };
 
         maybeAddProxyAuthorizationHeader(this.upstreamProxyUrlParsed, options.headers);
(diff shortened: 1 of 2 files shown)


regevbr:
Validated logic works in production

src/handler_tunnel_chain.js
```diff
@@ -20,12 +20,16 @@ export default class HandlerTunnelChain extends HandlerBase {
...
+
         const options = {
             method: 'CONNECT',
             hostname: this.upstreamProxyUrlParsed.hostname,
             port: this.upstreamProxyUrlParsed.port,
-            path: `${this.trgParsed.hostname}:${this.trgParsed.port}`,
-            headers: {},
+            path: targetHost,
+            headers: {
+                host: targetHost,
```
jancurn: This should be `Host`

regevbr: I will change though header names are case insensitive so that doesn't really matter

regevbr:
@jancurn sure I wlll do all that

src/handler_tunnel_chain.js
jancurn: Great, thanks. You're right, but it's typical to send the headers with the right case.

regevbr:
@jancurn done please review

regevbr:
@jancurn no problem! when can we expect a new version with the fix?

mtrunkat:
I have just released it as `0.2.6` with BETA tag.

regevbr:
@jancurn nice! already in production. Many thanks!

PR #47, opened 2019-09-20
Allow https scheme in upstream proxy url

dan-dr:
At the moment, when you return an `upsteramProxyUrl` with `https` scheme in `prepareRequestFunction`  you get an error: `Request failed with unknown error: Error: Invalid "upstreamProxyUrl" provided: URL must have the "http" scheme`

It's a shame that it only shows the error in verbose mode as it's pretty critical. Took me 30 minutes to figure it out because I was simply seeing `net::ERR_TUNNEL_CONNECTION_FAILED` in puppeteer.

Is there a reason for only checking `http`?
Also, let me know if the use of `Array.includes` is acceptable.

### DIFF
--- a/src/server.js
+++ b/src/server.js
@@ -299,8 +299,8 @@ export class Server extends EventEmitter {
                         if (!handlerOpts.upstreamProxyUrlParsed.hostname || !handlerOpts.upstreamProxyUrlParsed.port) {
                             throw new Error('Invalid "upstreamProxyUrl" provided: URL must have hostname and port');
                         }
-                        if (handlerOpts.upstreamProxyUrlParsed.scheme !== 'http') {
-                            throw new Error('Invalid "upstreamProxyUrl" provided: URL must have the "http" scheme');
+                        if (!['http', 'https'].includes(handlerOpts.upstreamProxyUrlParsed.scheme)) {
+                            throw new Error('Invalid "upstreamProxyUrl" provided: URL must have the "http" or "https" scheme');
                         }
                     }
                 }


jancurn:
Hi there, thanks for the PR. However, I'm not sure it's right to consider `https://` as a valid proxy URL for an HTTP proxy. If I'm not mistaken, HTTPS proxy is usually understood as a standard HTTP proxy to which the client connects via SSL, but that's something proxy-chain currently doesn't support. Therefore, supporting `https://` could actually cause confusion in what it actually does.

theArina:
@jancurn, don't you have plans to implement https feature ?

jancurn:
@theArina please can you explain which feature exactly you mean? What is your use case? How exactly it should work?

theArina:
@jancurn, sure, I ment the support of the https upstream proxy.

jancurn:
Please can you be more specific, what kind of HTTPS proxy you mean? proxy-chain supports HTTPS already, via HTTP CONNECT method. So do you mean tunneling HTTP CONNECT via SSL? Do you have any references to this

theArina:
@jancurn, well, first of all I get `proxy URL must have the 'http' protocol` error when I try to use an https proxy.

jancurn:
@theArina have you tried changing `https` to `http` to see if it works?

theArina:
@jancurn, so, it's only https proxy. Yes, I was taking about tunneling HTTP CONNECT via SSL.

theArina:
@jancurn, if you could give me some tips of how to implement this, I would see what I can do. We really need this feature in our project and your lib fits best for us.

PR #98, opened 2020-11-07
0 port support

kemphack:


src/server.js
```diff
@@ -93,7 +93,10 @@ export class Server extends EventEmitter {
 
         options = options || {};
 
-        this.port = options.port || DEFAULT_PROXY_SERVER_PORT;
+        if (options.port === undefined)
```
jancurn, line 96: Please can you also test for `null` ?

src/server.js
```diff
@@ -449,6 +452,7 @@ export class Server extends EventEmitter {
                 reject(err);
             };
             const onListening = () => {
+                this.port = this.server.address().port;
```
jancurn, line 455: Can we update some existing unit test to check that `this.port` remains set correctly to the previous value, if it was null, undefined or 0?

jancurn (changes requested):
Thanks for the PR, just two small comments, otherwise it's good to go.

src/server.js
kemphack: not sure, i added separate test for the first time

jancurn (approved):

PR #101, opened 2021-01-19
Fix closeTunnel function

andretunes:
The function `closeTunnel` was always returning `false`, since the `resolve` function was returning `undefined` and the function close of the running server that was mal formed, so it led to the tunnel not to close.


jancurn:
Good catch, thanks!

andretunes:
If there's anything else that I could help, just tell me :)

Have a nice day!

PR #102, opened 2021-01-19
Fix crash about Proxy Password

andretunes:
When Proxy Password has special characters it crashes and the function `url.parse()` is deprecated, so I modified to a constructor and now Proxy Password can have special characters.


jancurn:
Thanks for identifying the root cause, this has been fixed in https://github.com/apify/proxy-chain/pull/103 by @valekjo

PR #103, opened 2021-01-26
Improving parseUrl() to handle IPv6 and other special URLs

jancurn:
Actually, looking at it, it might be best to just get rid of `parseUrl` completely and use `new URL()`, it only adds `scheme` to the parsed object which can be easily replaced by `protocol` around the code base

### DIFF
--- a/src/tools.js
+++ b/src/tools.js
@@ -59,7 +59,8 @@ export const isInvalidHeader = (name, value) => {
  * Wraps `new URL(url)` and adds following:
  *  - `port` is casted to number / null from string
  *  - `path` field is added (pathname + search)
- *  - malformed or relative urls are parsed as well
+ *  - malformed or relative urls are parsed as well (always, the given string is returned as is in
+ *    few fields, other are left undefined)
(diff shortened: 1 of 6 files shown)


valekjo:
Sadly the `new URL` does more and there are some edge cases where it behaves slightly differently. The fix passes tests, parseUrl and redactUrl tests were updated.

src/tools.js
```diff
@@ -64,31 +64,34 @@ export const isInvalidHeader = (name, value) => {
...
+    };
 
-    if (parsed.auth) {
-        const matches = /^([^:]+)(:?)(.*)$/.exec(parsed.auth);
-        if (matches && matches.length === 4) {
-            parsed.username = matches[1];
-            if (matches[2] === ':') parsed.password = matches[3];
-        }
-    }
+    console.dir({ parsed });
```
jancurn: pls can we get rid of this?

jancurn (commented):
Looks good, thanks @valekjo

PR #109, opened 2021-02-12
Fix encoded passwords

### LINKED ISSUE #108: Support for passwords having ":" in upstream proxies
I'm currently using proxyland.io, they're using : in password to separate sessions
the https-proxy-agent module supports : in passwords,
would be really glad if this request is possible

mnmkng:
Fixes #108

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+0.4.10 / 2021-02-12
+===================
+- Fix percent encoded usernames and passwords causing an abrupt socket close.
+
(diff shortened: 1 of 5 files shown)


jancurn:
Thanks @mnmkng. But this was one of those PRs that make you realize that the core issue is much deeper and that it really needs a proper fix (e.g. the unit test was testing web server, not proxy auth) ... so I just fixed it, so thanks for nice Friday evening :)

So closing this PR in favor of  https://github.com/apify/proxy-chain/pull/110

PR #110, opened 2021-02-12
Fix the parseUrl() to correctly handle special chars in username or password

jancurn:
Plus other changes - see 0.5.0 in CHANGELOG.

These changes are potentially breaking, hence bumping major version.

Unfortunately, we don't have the CI running, so I don't know if this will work well on all Node versions...


metalwarrior665:
@jancurn I hope we are still releasing the proxy password fix also as a path version so it propagates to older builds?

metalwarrior665:
@mnmkng For context, Alan was able to fix the proxy password by migrating to 0.22.4. So it might have been unrelated or I don't know :)

jancurn:
> @jancurn I hope we are still releasing the proxy password fix also as a patch version so it propagates to older builds?

@metalwarrior665 the fix has increased major version number, so the depending packages need to upgrade to that version ... in 99.9% it's simple package upgrade with no code changes

jancurn:
> @mnmkng For context, Alan was able to fix the proxy password by migrating to 0.22.4. So it might have been unrelated or I don't know :)

Pls can you explain what do you mean? Version 0.22.4 of what?

mnmkng:
I think the package was broken by the move from `url.parse()` to `new URL()` in the first place. So I think if we publish this change as `5.0`, we should also release `4.10` with a revert back to `url.parse()`.

See the screenshot. There are changes in almost all properties. The move should have been a breaking change. Especially considering the user info.
image

mnmkng:
> Pls can you explain what do you mean? Version 0.22.4 of what?

Since the SDK uses `proxy-chain@^4.x.x`, the previous update broke old versions of SDK for some users.

jancurn:
> Since the SDK uses `proxy-chain@^0.4.x`, the previous releases of `proxy-chain` broke old versions of SDK for some users.

Shall we revert those changes back?

jancurn:
Oh I see, right, will do it

jancurn:
Done, version 0.4.7 published as latest:

Screen Shot 2021-02-13 at 12 33 25

PR #112, opened 2021-02-13
Add GitHub Actions release workflows

mnmkng:
Release guide is [in Notion](https://www.notion.so/apify/Release-Process-7f2b48c4120b4ba6b9a7b3cae6c86583).


mnmkng:
Ok, got it, need to replicate this config from Travis in GitHub Actions

```
addons:
    hosts:
        - localhost-test
```

jancurn (approved):
Thanks @mnmkng, the deployment looks a bit complicated, but let's see.

We can try to release version 0.5.0 with this new workflow. Or perhaps call it version 1.0?

mnmkng:
@jancurn yeah sadly actions have much more complex syntax than Travis.

But we need to fix the tests first. Dunno why they're failing now. Will check that on Monday.

mnmkng:
@jancurn Ok, so it's a DNS resolution issue for the non-existent hostname. Not sure yet what's a proper way to fix that.
image

jancurn:
> Ok, so it's a DNS resolution issue for the non-existent hostname. Not sure yet what's a proper way to fix that

@mnmkng The `non-existent-hostname` should test an invalid DNS resolution, not sure what is the problem, the pipeline breaks on that?

PR #114, opened 2021-02-18
fix: remove dependency on node's `_http_common` module

### LINKED ISSUE #50: Avoid requiring the _http_common module
Otherwise, the package is not usable from Electron. See https://github.com/apifytech/apify-js/issues/522

The module is only needed for `isInvalidHeader()` in https://github.com/apifytech/proxy-chain/blob/master/src/tools.js#L2

Double-check that the HTTP header validation is compatible with Node.js' header validation (they change HTTP parser in version 12 I think, which has slightly different behavior), perhaps in unit tests we could test against Node's internal methods too, just to be sure.

B4nan:
The checks is `isInvalidHeader` method are now backported from node's source.
Both methods were just a regexp checks, so it was trivial to use the very same
logic as in node.

The implementation changed a bit, `checkInvalidHeaderChar` is now implemented via regexp check,
and the `validateHeader` method was split into `validateHeaderName` and `validateHeaderValue`.

Both regexps are taken from the latest code from master. Also verified the exiting test in node,
but they seem to be aligned with the existing test suite (added only one additional assertion
for `undefined`).

Closes #50

### DIFF
--- a/src/tools.js
+++ b/src/tools.js
@@ -1,5 +1,3 @@
-import { _checkIsHttpToken, _checkInvalidHeaderChar } from '_http_common'; // eslint-disable-line
-
 const HOST_HEADER_REGEX = /^((([a-zA-Z0-9]|[a-zA-Z0-9][a-zA-Z0-9-]*[a-zA-Z0-9])\.)*([A-Za-z0-9]|[A-Za-z0-9][A-Za-z0-9-]*[A-Za-z0-9]))(:([0-9]+))?$/;
 
 /**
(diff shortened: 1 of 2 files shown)


src/tools.js
```diff
@@ -42,15 +40,35 @@ const HOP_BY_HOP_HEADERS_REGEX = new RegExp(`^(${HOP_BY_HOP_HEADERS.join('|')})$
...
 export const isHopByHopHeader = (header) => HOP_BY_HOP_HEADERS_REGEX.test(header);
 
+const TOKEN_REGEX = /^[\^_`a-zA-Z\-0-9!#$%&'*+.|~]+$/;
+
+/**
+ * Verifies that the given val is a valid HTTP token per the rules defined in RFC 7230
+ * @see https://tools.ietf.org/html/rfc7230#section-3.2.6
+ * @see https://github.com/nodejs/node/blob/8cf5ae07e9e80747c19e0fc04fad48423707f62c/lib/_http_common.js#L222
+ */
+export const isHttpToken = (val) => TOKEN_REGEX.test(val);
```
B4nan, line 50: maybe its better not to export those?

src/tools.js
```diff
@@ -42,15 +40,35 @@ const HOP_BY_HOP_HEADERS_REGEX = new RegExp(`^(${HOP_BY_HOP_HEADERS.join('|')})$
...
+const HEADER_CHAR_REGEX = /[^\t\x20-\x7e\x80-\xff]/;
+
+/**
+ * True if val contains an invalid field-vchar
+ *  field-value    = *( field-content / obs-fold )
+ *  field-content  = field-vchar [ 1*( SP / HTAB ) field-vchar ]
+ *  field-vchar    = VCHAR / obs-text
+ * @see https://github.com/nodejs/node/blob/8cf5ae07e9e80747c19e0fc04fad48423707f62c/lib/_http_common.js#L233
+ */
+export const isInvalidHeaderChar = (val) => HEADER_CHAR_REGEX.test(val);
```
B4nan, line 61: maybe its better not to export those?

src/tools.js
jancurn: Yeah, I'd skip the exports, no need for it..

jancurn (approved):
Looks great, thanks!

Issue #30, opened 2018-12-17
How to get headers from Connect Event


Benraay:
Hi I'm using your proxy-chain module for some it works very well with my previous proxy with the anonymize function  but I have a new one : ProxyMesh, 
I need to send some headers only to the proxy and not to the final website, how can I achieve this ?
need to send this header "X-ProxyMesh-Not-IP" 

also I need to get some headers from the proxy after the first call to "connect" on the proxy
I need to get the "X-ProxyMesh-IP"
find more info here: https://docs.proxymesh.com/article/7-request-response-headers#https

whole documentation : https://docs.proxymesh.com/article/7-request-response-headers

I suppose I need to use the prepareRequestFunction but I could not get it working !

I need this on HTTPS final sites  !! the proxy is http

Benraay:
here is a code sample using the raw node http module I hope this helps

`const http = require('http')
const https = require('https')
const username = 'username'
const password = 'password'
const auth = 'Basic ' + Buffer.from(username + ':' + password).toString('base64');

http.request({
     host: 'fr.proxymesh.com', // IP address of proxy server
     port: 31280, // port of proxy server
     method: 'CONNECT',
     path: 'api.ipify.org', // some destination,
     headers: {
         'Host' : "https://api.ipify.org",
         'Proxy-Authorization': auth
         },
     }).on('connect', (res, socket) => {
         if (res.statusCode === 200) { // connected to proxy server
             //get the ip used by proxymesh here
             console.log("IP : "+res.headers["x-proxymesh-ip"]);
             https.get({
                     host: 'api.ipify.org',
                     socket: socket, // using a tunnel
                     agent: false    // cannot use a default agent
                     path: '/' //put here the path ont the target site
             }, (res) => {
                 let chunks = []
                 res.on('data', chunk => chunks.push(chunk));
                 res.on('end', () => {
                         console.log('DONE', Buffer.concat(chunks).toString('utf8'))
                 })
             })
         }
     }).on('error', (err) => {
           console.error('error', err)
     }).end();`

Benraay:
Hi here,

after some investigation I could find where the CONNECT action is done

it's done inside handler_tunnel_chain.js on line 44
the onTrgRequestConnect function returns a response object with the headers I need.

is there an easy way to listen to that event, I don't see any emit on that one, could be great to add some option for emiting that event with headers

jancurn:
[...]

PR #116, opened 2021-02-22
Exposing the CONNECT response details of chained tunnels via a new event

starrify:
This is considered useful when some proxy providers may include useful information (e.g. for debugging) in the CONNECT response headers.

Implements #30.

This PR is based on the earlier commit in #31 (thanks @Benraay!) with these additional changes:
- Resolved conflicts between the earlier commit and the current `master` branch.
- Added a new test case for this very feature
- Slightly adjusted the callback interface in `listenConnectAnonymizedProxy`
- Renamed the event name (`onTrgRequestConnect` -> `tunnelConnectResponded`) for this may be accessed by users directly (via `server.on`).

### DIFF
--- a/src/anonymize_proxy.js
+++ b/src/anonymize_proxy.js
@@ -99,6 +99,6 @@ export const listenConnectAnonymizedProxy = (anonymizedProxyUrl, onTrgRequestCon
         onTrgRequestConnectCallback(false);
     }
     server.on('onTrgRequestConnect', ({ response, socket, head }) => {
-        onTrgRequestConnectCallback(response,socket,head);
+        onTrgRequestConnectCallback({ response,socket,head });
     });
 };
--- a/src/anonymize_proxy.js
+++ b/src/anonymize_proxy.js
@@ -86,3 +86,19 @@ export const closeAnonymizedProxy = (anonymizedProxyUrl, closeConnections, callb
         });
     return nodeify(promise, callback);
 };
+
+/**
+ * Add a callback on 'onTrgRequestConnect' Event in order to get headers from CONNECT tunnel to proxy
+ * Useful for some proxies that are using headers to send information like ProxyMesh
+ * @param anonymizedProxyUrl
+ * @param onTrgRequestConnectCallback
+ */
+export const listenConnectAnonymizedProxy = (anonymizedProxyUrl, onTrgRequestConnectCallback) => {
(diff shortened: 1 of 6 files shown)


Benraay:
Good news !
Never got the time to add tests thanks for going on !

src/anonymize_proxy.js
```diff
@@ -86,3 +86,19 @@ export const closeAnonymizedProxy = (anonymizedProxyUrl, closeConnections, callb
...
+/**
+ * Add a callback on 'tunnelConnectResponded' Event in order to get headers from CONNECT tunnel to proxy
+ * Useful for some proxies that are using headers to send information like ProxyMesh
+ * @param anonymizedProxyUrl
+ * @param tunnelConnectRespondedCallback
+ */
+export const listenConnectAnonymizedProxy = (anonymizedProxyUrl, tunnelConnectRespondedCallback) => {
+    const server = anonymizedProxyUrlToServer[anonymizedProxyUrl];
+    if (!server) {
+        tunnelConnectRespondedCallback(false);
```
jancurn, line 99: Would be better to return `null`, to avoid mixing of types (object vs boolean). And you probably need to return from the function.

src/anonymize_proxy.js
```diff
@@ -86,3 +86,19 @@ export const closeAnonymizedProxy = (anonymizedProxyUrl, closeConnections, callb
         });
     return nodeify(promise, callback);
 };
+
+/**
+ * Add a callback on 'tunnelConnectResponded' Event in order to get headers from CONNECT tunnel to proxy
+ * Useful for some proxies that are using headers to send information like ProxyMesh
+ * @param anonymizedProxyUrl
+ * @param tunnelConnectRespondedCallback
```
jancurn, line 94: Please could you describe the callback params here?

jancurn (changes requested):
Looks great, many thanks! I've just a small comment. Also, would be great if you could add a short description to README about this event and function.

starrify:
Hi @jancurn , thanks for the rapid response and the review opinions. I've had this PR updated accordingly by adding two new commits.

jancurn (approved):
Awesome, many thanks for a nice PR.

mnmkng (approved):
Nice, thanks!

thomas510111:
I think it would be a good idea to also pass the proxyChainId. Created a PR.
https://github.com/apify/proxy-chain/pull/173

PR #124, opened 2021-04-07
parseUrl: return standard ports for http(s) when port is not specified

### LINKED ISSUE #123: Anonymize proxy throw hostname and port error when they are present
Hi,

From version 0.4.8 and on I can't anonymize URLs like `http://proxy:proxy321@my.domain.com`, `http://proxy:proxy321@my.domain.com:80` due to a change on `/src/tools.js:parseUrl`. This is happening because `new URL(...)` omit the default protocol port causing the `anonymizeProxy(...)` to throw an error `Invalid "proxyUrl" option: the URL must contain both hostname and port.`.

```
> new URL('http://proxy:proxy321@my.domain.com:80')
URL {
  href: 'http://proxy:proxy321@my.domain.com/',
  origin: 'http://my.domain.com',
  protocol: 'http:',
  username: 'proxy',
  password: 'proxy321',
  host: 'my.domain.com',
  hostname: 'my.domain.com',
  port: '', // Empty port event when it is present
  pathname: '/',
  search: '',
  searchParams: URLSearchParams {},
  hash: ''
}
```

A possible fix here would be set the default http port on `anonymizeProxy` when `parsedProxyUrl.protocol === 'http:'` and port is empty. Thoughts?

Thank you.

valekjo:
Proxy url returns standard ports for few protocols (see https://nodejs.org/api/url.html#url_url_port) when no port is explicitly specified.

Removed tests which needlesly require the standard port
 - in `createTunnel`, only the parsed url is used in the end, so it should be ok
 - in `anonymizeProxy`, the raw (unparsed) url is passed to `Server` through `prepareRequestFunction`. There it also uses only the parsed version.

Resolves #123

### DIFF
--- a/src/tools.js
+++ b/src/tools.js
@@ -80,9 +80,13 @@ const bulletproofDecodeURIComponent = (encodedURIComponent) => {
 };
 
 // Ports returned by `parseUrl` when port is not explicitly specified.
+// Values are based on node docs: https://nodejs.org/api/url.html#url_url_port
 const STANDARD_PORTS_BY_PROTOCOL = {
+    'ftp:': 21,
     'http:': 80,
     'https:': 443,
+    'ws:': 80,
+    'wss:': 443,
 };
 
 /**
--- a/src/tools.js
+++ b/src/tools.js
@@ -79,6 +79,12 @@ const bulletproofDecodeURIComponent = (encodedURIComponent) => {
     }
 };
 
+// Ports returned by `parseUrl` when port is not explicitly specified.
+const STANDARD_PORTS_BY_PROTOCOL = {
+    'http:': 80,
+    'https:': 443,
+};
+
 /**
  * Parses a URL using Node.js' `new URL(url)` and adds the following features:
  *  - `port` is casted to number / null from string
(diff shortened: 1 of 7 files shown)


src/tools.js
```diff
@@ -79,6 +79,16 @@ const bulletproofDecodeURIComponent = (encodedURIComponent) => {
...
 
+// Ports returned by `parseUrl` when port is not explicitly specified.
+// Values are based on node docs: https://nodejs.org/api/url.html#url_url_port
+const STANDARD_PORTS_BY_PROTOCOL = {
+    'ftp:': 21,
+    'http:': 80,
+    'https:': 443,
+    'ws:': 80,
+    'wss:': 443,
+};
```
gutobortolozzo, line 90: @valekjo Do you think worth adding socks default port?

https://tools.ietf.org/html/rfc1928

src/tools.js
```diff
@@ -128,6 +138,11 @@ export const parseUrl = (url) => {
         }
     }
 
+    // Add default port based on protocol when no port is explicitly specified.
```
jancurn, line 141: Pls can we add mention to docs?

jancurn (approved):
Good job @valekjo

src/tools.js
valekjo: I don't think it's necessary, unless we decide to add all standard ports. This covers ports that are "stripped out" by `new URL(url)`. Is that ok?

jancurn (approved):

jancurn:
@valekjo oh, pls can you update the CHANGELOG and bump package.json version? Not sure what the deploy process is (cc @mnmkng )

mnmkng (approved):

src/tools.js
gutobortolozzo: Sounds good

PR #130, opened 2021-04-21
fix EventEmitter memory leak

0xGREG:
When the listener is running I get this error every second: `MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 error listeners added to [Socket]. Use emitter.setMaxListeners() to increase limit`

When I launched my app with `node --trace-warnings app.js` it showed that the stack trace is at line 284 in server.js. The server registers the error listener repeatedly so I added a small hack to remove every error listener before adding a new one.

There might be better ways to solve this but this is a solution for now.

### DIFF
--- a/src/server.js
+++ b/src/server.js
@@ -211,6 +211,7 @@ export class Server extends EventEmitter {
         // See https://github.com/apify/proxy-chain/issues/53
         // TODO: HandlerBase will also attach its own 'error' handler, we should only attach this one
         //  if HandlerBase doesn't do it, to avoid duplicate logs
+        socket.removeAllListeners('error'); // remove all listeners to prevent memory leak
         socket.on('error', (err) => {
             this.log(handlerOpts.id, `Source socket emitted error: ${err.stack || err}`);
         });


jancurn:
@0xGREG thanks for the analysis and the fix. Are you using HTTP or HTTPS connections? I'm thinking there might be a better way to do this.

0xGREG:
@jancurn I use HTTP connections.

mnmkng:
@jancurn what solution were you thinking?

jancurn:
I think this might be caused by the fact that HTTP sockets are recycled, and the handlers keep attaching. Hence it only happens for HTTP connections, not HTTPs. I'd prefer a little more directed fix, the suggested solution looks bit like a hotfix that might introduce some other issues. But didn't have time to look more into this...

mnmkng:
Yeah, we'll look into it when we can.

0xGREG:
Hey, any update on this?

mnmkng:
Maybe @szmarczak would like to take this next week?

szmarczak:
Sure.

0xGREG:
Hey, I don't know if you had time to check @szmarczak , but I think a proper solution would be check if our error listener was added already or no, and if no, add it. I don't know how to enumerate the listeners though, I've been looking for a function but I couldn't find one.

szmarczak:
> I think a proper solution would be check if our error listener was added already or no, and if no, add it.

Have you seen my PR?

0xGREG:
> Have you seen my PR?

Just did, thanks!

PR #133, opened 2021-05-07
added support for custom proxy headers with upstream proxy

lbosse:


lbosse:
Meant to PR on my local fork....

Issue #121, opened 2021-03-30
Proxy headers


theArina:
How can I add proxy headers to upstream proxy ?

jancurn:
I assume you'd like to add extra headers to the `HTTP CONNECT` method, right? Currently, this is not possible. See `HandlerTunnelChain` implementation, PRs welcome.

theArina:
@jancurn, I'll take a look at this, thanks.

dmitryshelomanov:
@theArina has solution ?

theArina:
@dmitryshelomanov, yeah, we made a [fork.](https://github.com/petr-vysotskiy/proxy-chain) take a look, but we also added some enrichments that we needed.

dmitryshelomanov:
@theArina any docs ?

theArina:
@dmitryshelomanov, for a fork ?

theArina:
@dmitryshelomanov, well, shortly you just kinda return this
`{ requestAuthentication: false, upstreamProxyUrl, proxyHeaders };`
instead of this
`{ requestAuthentication: false, upstreamProxyUrl };`
from the **prepareRequestFunction** 
(but not necessarily)

bliuchak:
Closing this as resolved.

PR #144, opened 2021-07-29
Add proxy headers, add support for the https proxy

theArina:
Sorry for the mess though.  
#121, #47 related.

### DIFF
--- a/src/handler_forward.js
+++ b/src/handler_forward.js
@@ -1,4 +1,5 @@
 import http from 'http';
+import tls from 'tls';
 import {
     isHopByHopHeader, isInvalidHeader, addHeader, maybeAddProxyAuthorizationHeader,
 } from './tools';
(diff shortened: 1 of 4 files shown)


mnmkng:
Could you clean this up a bit please? There are a lot of unrelated changes in this PR caused by old commits. I have no idea what this PR would do when it has 900 changed lines in the diff.

theArina:
@mnmkng, sure.

theArina:
@mnmkng, hey, should I've even made a PR into this branch ? I now realise that it was updated long ago. And I was pulling changes from master.

mnmkng:
Yeah, you're right. I missed it. Please make the PR against master, not develop. Thanks.

theArina:
@mnmkng, here you go!

jancurn:
@theArina thanks for your PR and the cleanup. However, there are still zero comments explaining what your changes do and what they are good for, and zero new unit tests for those changes. Also, I don't think it makes sense to add one more dependency "rimraf" just to enable "npm clean" command, that's just too wasteful.

B4nan:
> Also, I don't think it makes sense to add one more dependency "rimraf" just to enable "npm clean" command, that's just too wasteful.

It depends - if we want to make this repository windows friendly, we can't use just `rm` (that's the point of `rimraf`, being platform independent). But it should be a **dev dependency**.

Granted, this is not related to the PR itself and would be better done separately.

theArina:
I'll see what I can do with the comments and stuff. 
It's just at first we made changes for the personal usage, but I thought someone else might also need this stuff.

jancurn:
@theArina I totally understand and are grateful for your help, we just require a certain standard for this library, as it's used quite a bit, including our own production systems.

src/handler_forward.js
```diff
@@ -16,9 +17,104 @@ export default class HandlerForward extends HandlerBase {
...
+            host: this.upstreamProxyUrlParsed.hostname,
+            port: this.upstreamProxyUrlParsed.port
+        });
+
+        const headers = { ...this.proxyHeaders };
+        maybeAddProxyAuthorizationHeader(this.upstreamProxyUrlParsed, headers);
+        const requestUrl = new URL(this.srcRequest.url);
+        const hostname = `${requestUrl.hostname}:${requestUrl.port || '80'}`;
+
+        let payload = `CONNECT ${hostname} HTTP/1.1\r\n`;
```
szmarczak, line 38: Using the CONNECT protocol is unnecessary. This is just a regular web proxy.

https://github.com/apify/proxy-chain/blob/0043ac9db2c04fffdd53fe7ee7d3a11183648555/src/handler_forward.js#L114

```js
const fn = reqOpts.protocol === 'https:' ? https.request : http.request;
this.trgRequest = fn(reqOpts); 
```

and it should work.

szmarczak: Ah, I just realized the upstream proxy server doesn't have to support HTTPS...

PR #146, opened 2021-08-12
Fix event emitter memory leak

szmarczak:
Fixes #130

The issue is that `proxy-chain` assumes each request is a new socket (`HandlerBase`). Unfortunately in case of `keep-alive`, this isn't true, so the events keep stacking. No matters if this is HTTP or HTTPS. The memory leak occurs only when there are multiple handler for the same socket - IOW those are **not** `CONNECT` requests but regular HTTP/1.1 methods such as `GET`.

This doesn't fix the warnings but fixes the memory leaks. Also I'm not sure if this is the right solution, the logic seems pretty complicated 😅 Best if we opreated only on `req` and `res` for regular HTTP/1.1 requests.


src/handler_base.js
```diff
@@ -73,6 +73,26 @@ export default class HandlerBase extends EventEmitter {
         this.srcSocket.on('error', this.onSrcSocketError);
     }
 
+    detach() {
```
jancurn, line 76: Can we please add a comment to explain what this function does and what is the difference from `close` ? And do we even need to add this function? Maybe we could just add the `off()` calls to `close` and skip this function altogether?

src/handler_forward.js
```diff
@@ -154,6 +154,10 @@ export default class HandlerForward extends HandlerBase {
 
         this.srcResponse.writeHead(response.statusCode, headers);
         response.pipe(this.srcResponse);
+
+        response.once('end', () => {
```
jancurn, line 158: Comment here too perhaps?

jancurn (commented):
Many thanks @szmarczak for looking into this! Please see my comments

src/handler_forward.js
pocesar: is there any situation where this event listener is attached more than once and never triggered, since the `end` event is [notoriously flakey](https://nodejs.org/api/stream.html#stream_event_end)? might lead to the MaxListenersExceededWarning

src/handler_base.js
szmarczak: `close` calls `socket.destroy()` and we don't want that. Or do we? If we do then we won't support `keep-alive` but only `close`.

src/handler_forward.js
szmarczak: @pocesar If this isn't called then `error` will which will happen to destroy the socket.

test/ee-memory-leak.js
```diff
@@ -0,0 +1,61 @@
...
+
+        let socket;
+        let registeredCount;
+        proxyServer.server.prependOnceListener('request', (request) => {
+            socket = request.socket;
+            registeredCount = socket.listenerCount('error');
+        });
+
+        const callback = () => {
+            console.log(socket.listenerCount('error'), registeredCount);
```
mnmkng, line 33: forgotten console?

mnmkng (approved):

test/ee-memory-leak.js
szmarczak:
```suggestion
```

src/handler_base.js
jancurn: But if we just detach the listeners here, the pipe between `srcSocket` and `trgSocket` will remain open, and the data will keep flowing, no? That's something we don't want as reusing such socket for another connection will cause other issues. In theory we could keep the `srcSocket` open, shut down the pipe, and close the `trgSocket`, or even reuse the pipe, but I'm not sure it's worth the effort for HTTP connections only, as they comprise only a small fraction of traffic.

Maybe shutting down the sockets (after the data from target to source is surely sent) is the easiest option, even though `keep-alive` would be ignored.

szmarczak: > the pipe between srcSocket and trgSocket will remain open, and the data will keep flowing, no?

In case of `handler_forward` there is no such pipe. The pipe is between `srcRequest` and `trgRequest` + `response` (trgResponse) and `srcResponse`. `detach()` is called only on resposne 'end' event. Therefore the pipe will be closed and no data will be emitted.

> In theory we could keep the srcSocket open, shut down the pipe, and close the trgSocket, or even reuse the pipe, but I'm not sure it's worth the effort for HTTP connections only, as they comprise only a small fraction of traffic.

To be honest I feel like `handler_base` is a bit too abstract. It tries to accept both tunnels and usual requests, which have different APIs under the hood. `proxy-chain` was created 3 yrs ago, so that could be best solution at that time.

> Maybe shutting down the sockets (after the data from target to source is surely sent) is the easiest option, even though keep-alive would be ignored.

I agree. Let me fix this quick.

szmarczak: > Maybe shutting down the sockets (after the data from target to source is surely sent) is the easiest option, even though keep-alive would be ignored.

Actually this still doesn't fix the issue because there is no option to disable pipelining in Node.js HTTP server.

szmarczak: I can move `detach` to `handler_forward` as it's used only there. Wdyt?

jancurn: thanks @szmarczak this sounds reasonable

szmarczak: Moved the `detach` function to `handler_forward.js` :) Feel free to merge now.

jancurn (approved):
Awesome, thanks @szmarczak !

src/handler_forward.js
```diff
@@ -154,11 +154,43 @@ export default class HandlerForward extends HandlerBase {
...
     onTrgError(err) {
         if (this.isClosed) return;
         this.log(`Target socket failed: ${err.stack || err}`);
         this.fail(err);
     }
+
+    /**
+     * Detach removes all listeners registered by HandlerBase.
+     * Must be called when the handler finishes on success,
+     * in order to prevent event emitter memory leak.
```
jancurn, line 175: Perhaps we could also add a link to https://github.com/apify/proxy-chain/pull/130

szmarczak: done 👍🏼

src/handler_forward.js
```diff
@@ -154,11 +154,44 @@ export default class HandlerForward extends HandlerBase {
...
     }
 
     onTrgError(err) {
         if (this.isClosed) return;
         this.log(`Target socket failed: ${err.stack || err}`);
         this.fail(err);
     }
+
+    /**
+     * @see https://github.com/apify/proxy-chain/pull/130
```
szmarczak, line 173:
```suggestion
     * @see https://github.com/apify/proxy-chain/issues/81
     * @see https://github.com/apify/proxy-chain/pull/130
```

szmarczak:
@jancurn I think it's good to merge now :)

PR #153, opened 2021-09-07
wip: simplify forward

szmarczak:
- [x] remove babel
- [x] added https support
  - [ ] test
- [x] require node.js 14

### TODO (those will be separate PRs)

- [ ] simplify direct
- [ ] simplify chain
- [ ] simplify custom response
- [ ] simplify server
- [ ] simplify tools
- [ ] use typescript


src/anonymize_proxy.js
```diff
@@ -1,7 +1,7 @@
-import { Server } from './server';
-import {
+const { Server } = require('./server');
```
B4nan, line 1: does it really make sense to do esm -> cjs conversion just to get rid of babel, when we plan to rewrite to ts with esm in the following step?

szmarczak: babel didn't like node.js streams for some reason. it called some function that did not exist (don't remember exactly what function was being called). maybe it needed more configuration, for me it was just easier to drop babel. only es imports were used.

`proxy-chain` doesn't use esm yet https://cdn.jsdelivr.net/npm/proxy-chain@1.0.4/build/index.js

B4nan: i am talking about the source code, it uses esm style imports. and once we make this TS, we will use them again. not talking about what gets compiled/published.

but sure, if there were real issue with babel, lets get rid of it

szmarczak: > i am talking about the source code, it uses esm style imports. and once we make this TS, we will use them again.

that's right.

> if there were real issue with babel, lets get rid of it

it generated something like this

```js
    static forward(request, response) {
        return _asyncToGenerator(function* () {
            var pipeline = util.promisify(stream.pipeline);

            yield pipeline(got.stream(request.url, {
                method: request.method,
                headers: request.headers,
                decompress: false,
                followRedirect: false
            }), response);
        })();
    }
```

and `_asyncToGenerator` was undefined.

test/server.js
```diff
@@ -547,8 +549,8 @@ const createTestSuite = ({
                     return requestPromised(opts)
                         .then((response) => {
                             if (useMainProxy) {
-                                expect(response.statusCode).to.eql(500);
-                                expect(response.body).to.match(/with an invalid HTTP status code/);
+                                expect(response.statusCode).to.eql(502);
+                                expect(response.body).to.eql('Bad status!');
```
szmarczak, line 553: Now instead of rejecting the entire request on invalid status code, the body will be reused and the status code will be set to 502.

src/forward.js
```diff
@@ -0,0 +1,108 @@
+const http = require('http');
+const stream = require('stream');
+const util = require('util');
+const { isHopByHopHeader } = require('./tools');
+
+const pipeline = util.promisify(stream.pipeline);
+
+const validHeadersOnly = (array) => {
```
szmarczak, line 8: This should be moved to utils I think

package.json
```diff
@@ -40,9 +40,7 @@
     "node": ">=10"
   },
   "dependencies": {
-    "got-scraping": "^3.2.3",
-    "underscore": "^1.9.1",
-    "undici": "^4.5.1"
+    "underscore": "^1.9.1"
```
B4nan, line 43: would be good if we get rid of this too, just peeked at the usage and should be easy

szmarczak: done

src/forward.js
jancurn: Any chance to add comments?

src/forward.js
```diff
@@ -0,0 +1,111 @@
...
+        }
+
+        rawHeaders.push(name, value);
+    }
+
+    return rawHeaders;
+};
+
+// eslint-disable-next-line no-async-promise-executor
+const forward = async (request, response, handlerOpts) => new Promise(async (resolve, reject) => {
```
jancurn, line 44: Please can you add comments here please what the function does, and what are the params? Perhaps also describe why this is not inherited from HandlerBase, it's not immediately obvious.

jancurn (commented):
Looks good, I just have a few comments.

src/forward.js
szmarczak: > Perhaps also describe why this is not inherited from HandlerBase, it's not immediately obvious.

My idea is that each handler will be a function like this, there's no need for `HandlerBase` at all.

jancurn: Yeah, just pls comment what and why :)

src/forward.js
szmarczak: Added comments, ptal :)

PR #156, opened 2021-09-13
wip: simplify chain

szmarczak:


src/server.js
```diff
@@ -179,11 +179,11 @@ class Server extends EventEmitter {
 
                 if (handlerOpts.upstreamProxyUrlParsed) {
                     this.log(handlerOpts.id, 'Using HandlerTunnelChain');
-                    return this.handlerRun(new HandlerTunnelChain(handlerOpts));
+                    return chain(request, socket, head, handlerOpts, this);
                 }
 
                 this.log(handlerOpts.id, 'Using HandlerTunnelDirect');
-                return direct(request, socket, head);
+                return direct(request, socket, head, handlerOpts, this);
```
B4nan, line 186: looks like there are just the 3 parameters on `direct`?

szmarczak: Fixed.

szmarczak:
```
  1) tcp_tunnel.createTunnel
       creates tunnel that is able to transfer data:
     Uncaught Error: listen EADDRINUSE: address already in use :::38102
      at Server.setupListenHandle [as _listen2] (node:net:1319:16)
      at listenInCluster (node:net:1367:12)
      at Server.listen (node:net:1454:7)
      at /home/runner/work/proxy-chain/proxy-chain/test/tcp_tunnel.js:19:12
      at new Promise (<anonymous>)
      at serverListen (test/tcp_tunnel.js:18:40)
      at /home/runner/work/proxy-chain/proxy-chain/test/tcp_tunnel.js:154:24
      at tryCatcher (node_modules/bluebird/js/main/util.js:26:23)
      at Promise._settlePromiseFromHandler (node_modules/bluebird/js/main/promise.js:510:31)
      at Promise._settlePromiseAt (node_modules/bluebird/js/main/promise.js:584:18)
      at Async._drainQueue (node_modules/bluebird/js/main/async.js:128:12)
      at Async._drainQueues (node_modules/bluebird/js/main/async.js:133:10)
      at Immediate.Async.drainQueues (node_modules/bluebird/js/main/async.js:15:14)
      at processImmediate (node:internal/timers:464:21)
```

@mnmkng I just got the error @rubydev saw before, looks like Node.js selected a protocol that was already in use?

src/server.js
B4nan: now it has 4 params, and you are feeding 5 (and server is 5th, not 4th), right?

szmarczak: Ah, this one :) Now it's 5, `handlerOpts` won't be used but I prefer to keep this one so it looks the same like `chain`.

B4nan (approved):
idea for future PRs - maybe we should turn the 5 parameters into object

mnmkng:
> https://github.com/apify/proxy-chain/runs/3586133160
> 
> 
> 
> ```
> 
>   1) tcp_tunnel.createTunnel
> 
>        creates tunnel that is able to transfer data:
> 
>      Uncaught Error: listen EADDRINUSE: address already in use :::38102
> 
>       at Server.setupListenHandle [as _listen2] (node:net:1319:16)
> 
>       at listenInCluster (node:net:1367:12)
> 
>       at Server.listen (node:net:1454:7)
> 
>       at /home/runner/work/proxy-chain/proxy-chain/test/tcp_tunnel.js:19:12
> 
>       at new Promise (<anonymous>)
> 
>       at serverListen (test/tcp_tunnel.js:18:40)
> 
>       at /home/runner/work/proxy-chain/proxy-chain/test/tcp_tunnel.js:154:24
> 
>       at tryCatcher (node_modules/bluebird/js/main/util.js:26:23)
> 
>       at Promise._settlePromiseFromHandler (node_modules/bluebird/js/main/promise.js:510:31)
> 
>       at Promise._settlePromiseAt (node_modules/bluebird/js/main/promise.js:584:18)
> 
>       at Async._drainQueue (node_modules/bluebird/js/main/async.js:128:12)
> 
>       at Async._drainQueues (node_modules/bluebird/js/main/async.js:133:10)
> 
>       at Immediate.Async.drainQueues (node_modules/bluebird/js/main/async.js:15:14)
> 
>       at processImmediate (node:internal/timers:464:21)
> 
> ```
> 
> 
> 
> @mnmkng I just got the error @rubydev saw before, looks like Node.js selected a protocol that was already in use?

@szmarczak That's weird. Could there be race condition somewhere?

jancurn:
> That's weird. Could there be race condition somewhere?

In the past we used to pick randomized ports, which was a race condition. Now it should be done by operating system, hopefully without races... so this is strange, unless this is an error from an older version?

jancurn:
Looking at all these PRs, I'm increasingly more worried that this big rewrite can potentially bring a lot of problems to this module that we rely on so heavily. There were a lot of iterations and a lot of real-world testing that went into this module to make sure it was stable and performing well (in particular the socket lifecycle), and now I feel we might be back to square one and might have to go through all the (hard-to-find) errors again, potentially with a big business impact (think of all potential problems a buggy Apify Proxy can cause to our customers).

I better stop watching these PRs and I truly hope you guys know what you're doing. Please make sure to write really good comments for the functions, so that other people can understand what's going on.

src/chain.js
```diff
@@ -0,0 +1,88 @@
...
+        port: proxy.port,
+        path: request.url,
+        headers: [
+            'host',
+            request.url,
+        ],
+    };
+
+    if (proxy.username || proxy.password) {
+        const auth = `${proxy.username}:${proxy.password}`;
```
jancurn, line 29: What is either `username` or `password` is null? Then the auth string will contain "null".

szmarczak: In WHATWG URL it's never null, it's an empty string.

szmarczak:
> Could there be race condition somewhere?

The tests still use 

https://github.com/apify/proxy-chain/blob/0043ac9db2c04fffdd53fe7ee7d3a11183648555/test/tools.js#L18-L34

so I guess it generated an in-use port. I'm starting to think @rubydev's error is unrelated to this one.

> now I feel we might be back to square one and might have to go through all the (hard-to-find) errors again

Currently all the tests pass (except the two ones with stats, which will be fixed later). The point of the rewrite is so we don't have to mutate the socket if there's no need to. Only operate on objects we really have to. So if we receive a normal HTTP/1.1 request, let's operate only on `request` and `response` objects, no need to tinker with `socket`. This way we'll prevent event emitter memory leaks and the code will be more readable.

> There were a lot of iterations and a lot of real-world testing that went into this module to make sure it was stable and performing well

We can gradually roll out the rewrite. Initially e.g. 1% of all requests can go through the rewrite, the next day it can be 5%, the next day it can be 10% etc.

> in particular the socket lifecycle

What about it? As I see it, it remains the same. The socket gets closed only when a user does so or there's an error on any side. Am I missing something? If you have any questions, I'm happy to answer them.

> I better stop watching these PRs and I truly hope you guys know what you're doing.

I'd prefer if you reviewed the code, you are the creator, but I'm not insisting. I'm trying to keep the code as simple as possible.

src/chain.js
szmarczak: With TypeScript we won't have to worry `can this be null?` :)

szmarczak:
I've added `more descriptive errors` on my todo list. So if an error gets thrown, we'll know all the details, upstream proxy, status code, request options etc.

jancurn:
Thanks @szmarczak it was just a bit of an old man's afraidness of change that was speaking :) Of course, I'll have a look at the PRs, I'll do a deeper review of the big PR once it's ready. I'm confident your changes will make the module better and cleaner. Just pls comment the functions well, to make it easier to review.

Regarding the socket life-cycle, there were some rare strange races that "should never happen" but indeed happened, which led to code blocks like:

```
// See https://github.com/apify/proxy-chain/issues/63
if (this.isClosed) return;
```

These come from battle-testing, so it might happen they will reoccur. But let's see... 

The gradual roll-out is a good idea, once this is ready pls chat with the Platform team how to do it.

szmarczak:
> it was just a bit of an old man's afraidness of change that was speaking :)

I understand the worry :) 

> I'll have a look at the PRs, I'll do a deeper review of the big PR once it's ready.

Thanks a lot ❤️ Two heads always better than one :D

> Just pls comment the functions well, to make it easier to review.

👍🏼 

> which led to code blocks like

This is also what I'm trying to prevent. `HandlerBase` created a response on top of a detached socket (from CONNECT method) from the native `http` module:

https://github.com/apify/proxy-chain/blob/0043ac9db2c04fffdd53fe7ee7d3a11183648555/src/handler_base.js#L43-L49

therefore there were more critical points that needed to be handled.

szmarczak:
```
// See https://github.com/apify/proxy-chain/issues/63
if (this.isClosed) return;
```

There's a race condition. Before the end socket gets connected the client one to the proxy may get destroyed (e.g. via disconnect).

mnmkng (approved):

PR #155, opened 2021-09-13
wip: simplify direct

szmarczak:
#### this PR

- [x] simplify direct

#### [wip: simplify forward](https://github.com/apify/proxy-chain/pull/153)

- [x] remove babel
- [x] simplify forward handler
- [x] added https support
  - [ ] test
- [x] require node.js 14
- [ ] skipped 2 tests (will re-enable once proxy-chain is fully rewritten)

#### [wip: simplify custom response](https://github.com/apify/proxy-chain/pull/154)

- [x] simplify custom response
- [x] use shared eslint config

#### TODO (those will be separate PRs)

- [ ] simplify chain
- [ ] simplify server
- [ ] simplify tools
- [ ] use typescript

PR #154, opened 2021-09-13
wip: simplify custom response

szmarczak:
#### this PR

- [x] simplify custom response
- [x] use shared eslint config

#### [wip: simplify forward](https://github.com/apify/proxy-chain/pull/153)

- [x] remove babel
- [x] simplify forward handler
- [x] added https support
  - [ ] test
- [x] require node.js 14
- [ ] skipped 2 tests (will re-enable once proxy-chain is fully rewritten)

#### TODO (those will be separate PRs)

- [ ] simplify direct
- [ ] simplify chain
- [ ] simplify server
- [ ] simplify tools
- [ ] use typescript

PR #157, opened 2021-09-14
wip: simplify tools

szmarczak:
#### this PR

- [x] simplify tools
  - [x] remove `parseUrl(...)` in favor of `new URL(...)`
  - [x] remove `parseHostHeader(...)` in favor of `new URL(...)`
  - [x] remove `isInvalidHeader(...)` (unused)
  - [x] tests: remove buggy `findFreePort` in favor of `port: 0`
  - [x] simplify `redactUrl`
  - [x] remove `addHeader` (leftover from #156)
  - [x] ~~ignore invalid `:` in username fields (if present, the upstream proxy will reject with not authorized error)~~
- [x] add IPv6 tests
- [x] revert `ignore invalid :`

There's a slight behavior change in `parseAuthorizationHeader`. It now returns `{ type: '', data: '' }` (instead of `null`) if the `proxy-authorization` header is invalid.

#### [wip: simplify forward](https://github.com/apify/proxy-chain/pull/153)

- [x] remove babel
- [x] added https support
  - [ ] test
- [x] require node.js 14

#### [wip: simplify custom response](https://github.com/apify/proxy-chain/pull/154)

- [x] simplify custom response
- [x] use shared eslint config

#### [wip: simplify direct](https://github.com/apify/proxy-chain/pull/155)

- [x] simplify direct

#### [wip: simplify chain](https://github.com/apify/proxy-chain/pull/156)

- [x] simplify chain
  - [x] remove handler base

#### TODO (those will be separate PRs)

- [ ] simplify server
- [ ] use typescript
- [ ] turn the 5 parameters of `direct` and `chain` into object
- [ ] more descriptive errors
- [ ] enable coverage
- [ ] simplify `TcpTunnel` (needed for https://blog.apify.com/tunneling-arbitrary-protocols-over-http-proxy-with-static-ip-address-b3a2222191ff/ )


test/tools.js
```diff
@@ -35,174 +35,6 @@ const findFreePort = () => {
         });
 };
 
-const testUrl = (url, expected) => {
```
jancurn, line 38: These unit tests covered plenty of situations (e.g. IPv6 and other special URLs) that are not covered by the main unit tests. Can we add those tests?

src/server.js
```diff
@@ -236,15 +235,11 @@ class Server extends EventEmitter {
 
                     let parsed;
                     try {
-                        parsed = parseUrl(request.url);
+                        parsed = new URL(request.url);
                     } catch (e) {
                         // If URL is invalid, throw HTTP 400 error
                         throw new RequestError(`Target "${request.url}" could not be parsed`, 400);
                     }
-                    // If srcRequest.url is something like '/some-path', this is most likely a normal HTTP request
```
jancurn, line 244: This error was supposed to be helpful for debugging...

mnmkng:
I'm not sure about the removal of `parseUrl`. @szmarczak please check some of the latest closed issues and the same for PRs. We were doing quite a lot of changes to the URL parsing logic recently for various reasons. I don't remember them all, but it should all be documented here on GitHub.

src/server.js
mnmkng: Agree. I liked this error 😄

szmarczak: This never executed because it throws `Target "/" could not be parsed` with `400`.

test/tools.js
szmarczak: This PR is using WHATWG URLs directly. Can you elaborate please?

szmarczak:
I also noticed that `TcpTunnel` is completely unused. Can we remove this?

mnmkng:
> I think you're having this in mind,

Yup, please also check all the related PRs and issues for potential problems :)

szmarczak:
There's still some things to do here but this is ready for review.

src/server.js
mnmkng: Actually I saw it in the wild multiple times. Not too long ago even.

szmarczak:
```
$ node src/run_locally.js
Proxy server is running at http://127.0.0.1:8080
```

image

jancurn:
> I also noticed that `TcpTunnel` is completely unused. Can we remove this?

Isn't this used for e.g. https://blog.apify.com/tunneling-arbitrary-protocols-over-http-proxy-with-static-ip-address-b3a2222191ff/ ?

szmarczak:
Ah. `TcpTunnel` can be simplified as well. I'll add this to my todo list.

src/tcp_tunnel_tools.js
```diff
@@ -1,11 +1,11 @@
...
 const TcpTunnel = require('./tcp_tunnel');
-const { parseUrl, nodeify } = require('./tools');
+const { nodeify } = require('./utils/nodeify');
 
 const runningServers = {};
 
 function createTunnel(proxyUrl, targetHost, providedOptions = {}, callback) {
-    const parsedProxyUrl = parseUrl(proxyUrl);
+    const parsedProxyUrl = new URL(proxyUrl);
     if (!parsedProxyUrl.hostname || !parsedProxyUrl.port) {
```
B4nan, line 9: looking at the `parseUrl`, it automatically added default ports, while `new URL` will keep an empty string if port is not explicitly provided, so I'd say this validation can be triggered for portless urls now?

test/tcp_tunnel.js
```diff
@@ -56,7 +54,7 @@ describe('tcp_tunnel.createTunnel', () => {
     it('throws error if proxyUrl is not in correct format', () => {
         assert.throws(() => { createTunnel('socks://user:password@whatever.com:123', 'localhost:9000'); }, /must have the "http" protocol/);
         assert.throws(() => { createTunnel('socks5://user:password@whatever.com', 'localhost:9000'); }, /must contain hostname and port/);
-        assert.throws(() => { createTunnel('http://us%3Aer:password@whatever.com:345', 'localhost:9000'); }, /cannot contain the colon/);
+        // assert.throws(() => { createTunnel('http://us%3Aer:password@whatever.com:345', 'localhost:9000'); }, /cannot contain the colon/);
```
B4nan, line 57: we should probably add a TODO here and reenable the test later?

B4nan:
btw would be good to fix the linting issues separately (separate PR, not just commit) to ease the reviewing a bit

src/tcp_tunnel_tools.js
szmarczak: `TcpTunnel` will be another PR

test/tcp_tunnel.js
szmarczak: >  ignore invalid : in username fields (if present, the upstream proxy will reject with not authorized error)

src/server.js
mnmkng: Still, I'm 100% sure I saw it in the wild. You don't imagine things like "I'm a HTTP proxy and not a regular web server" in error messages 😄 So there must've been some workflow where this actually triggered.

src/anonymize_proxy.js
```diff
@@ -15,8 +13,8 @@ const anonymizedProxyUrlToServer = {};
  * anonymous proxy URL or the original URL if it was already anonymous.
  */
 const anonymizeProxy = (proxyUrl, callback) => {
-    const parsedProxyUrl = parseUrl(proxyUrl);
-    if (!parsedProxyUrl.host || !parsedProxyUrl.port) {
+    const parsedProxyUrl = new URL(proxyUrl);
+    if (!parsedProxyUrl.hostname || !parsedProxyUrl.port) {
```
mnmkng: When we change the parsing logic, I'm pretty sure that it will no longer work for some users so we need to keep in mind that we should only update `proxy-chain` with a major update of `browser-pool` and `apify-js`. cc @B4nan

src/server.js
szmarczak: > Still, I'm 100% sure I saw it in the wild.

Perhaps an outdated `proxy-chain` server? This is not possible even with master as `parseUrl` calls `new URL` and `new URL('/')` always throws, so the `if` is never reached.

src/anonymize_proxy.js
B4nan, line 17: this is the same as i noted in the tcp tunnel change, right? port can be empty string and the validation will be triggered

src/tcp_tunnel_tools.js
B4nan: but you broke it in this one, right? so i guess at least add a todo

src/server.js
```diff
@@ -298,19 +291,18 @@ class Server extends EventEmitter {
...
-                        throw new Error(`Invalid "upstreamProxyUrl" provided: URL must have hostname and port (was "${funcResult.upstreamProxyUrl}")`);
-                    }
                     if (handlerOpts.upstreamProxyUrlParsed.protocol !== 'http:') {
                         // eslint-disable-next-line max-len
                         throw new Error(`Invalid "upstreamProxyUrl" provided: URL must have the "http" protocol (was "${funcResult.upstreamProxyUrl}")`);
                     }
+                    if (!handlerOpts.upstreamProxyUrlParsed.hostname) {
+                        // eslint-disable-next-line max-len
+                        throw new Error(`Invalid "upstreamProxyUrl" provided: URL must have hostname (was "${funcResult.upstreamProxyUrl}")`);
+                    }
```
mnmkng, line 305: Why do we no longer expect port?

src/tools.js
```diff
@@ -1,341 +0,0 @@
...
-    };
-
-    // Add scheme field (as some other external tools rely on that)
-    if (parsed.protocol) {
-        const matches = /^([a-z0-9]+):$/i.exec(parsed.protocol);
-        if (matches && matches.length === 2) {
-            // eslint-disable-next-line prefer-destructuring
-            parsed.scheme = matches[1];
-        }
-    }
```
mnmkng, line 148: I understand why you want to remove it, but I hope it won't cost us large refactorings somewhere in the core codebase.

test/tcp_tunnel.js
B4nan: i dont follow. you are commenting out an assertion, if its valid and you plan to uncomment it, please add a todo so we don't forget about that. if its invalid, why should we keep it commented instead of removing it?

src/tools.js
```diff
@@ -1,341 +0,0 @@
...
-// Replacement for Bluebird's Promise.nodeify()
-const nodeify = (promise, callback) => {
-    if (typeof callback !== 'function') return promise;
-
-    const p = promise.then((result) => callback(null, result), callback);
-
-    // Handle error from callback function
-    p.catch((e) => {
-        setTimeout(() => { throw e; }, 0);
-    });
```
mnmkng, line 292: Not sure about removing this. Without it, the code produces unhandled rejections. With it, it produces uncaught exceptions. It's semantics, but they are handled differently. Since the function is `nodeify` and should produce a callback interface, it should not produce rejections when the callback throws because that's not how a callback interface behaves. You also should not be able to catch the error downstream by adding a catch handler to the promise later.

On the other hand, I have no idea how important it is for the package, it just caught my eye.

src/utils/valid_headers_only.js
```diff
@@ -0,0 +1,46 @@
+const { validateHeaderName, validateHeaderValue } = require('http');
+const { isHopByHopHeader } = require('./is_hop_by_hop_header');
+
+/**
+ * Filters out invalid headers.
+ * @param {string[]} array
+ * @returns Filtered headers.
+ */
+const validHeadersOnly = (array) => {
+    const rawHeaders = [];
```
mnmkng, line 10: Those moves to a separate file could've probably been done in a separate PR. It's hard to see if there are any differences in the function code this way.

test/anonymize_proxy.js
```diff
@@ -159,7 +158,9 @@ describe('utils.anonymizeProxy', function () {
                 ]);
             })
             .then((results) => {
+                // eslint-disable-next-line prefer-destructuring
                 proxyUrl1 = results[0];
+                // eslint-disable-next-line prefer-destructuring
```
mnmkng, line 163: Maybe `[proxyUrl1, proxyUrl2] = results;` instead of the comments?

test/tcp_tunnel.js
mnmkng: > ignore invalid ...

So the behavior will be different? I think we should keep track of all the breaking and possibly breaking changes so that we can compile a good changelog.

src/tcp_tunnel_tools.js
szmarczak: The todo is in the PR description, but fine, I'll remove it (it's moot since the protocol check is enough).

This is the wrong PR to fix things like this.

src/server.js
szmarczak: We do. It's already covered by the protocol check. It's an empty string if it's a default port.

src/anonymize_proxy.js
szmarczak: > When we change the parsing logic, I'm pretty sure that it will no longer work

Why wouldn't it work? This check is removed now, the protocol check is sufficient.

src/tools.js
szmarczak: Hmmm... You're right. This doesn't crash the process on Node.js 14, but does on 16. I'll revert this change.

src/utils/valid_headers_only.js
szmarczak: [`f04773e` (#157)](https://github.com/apify/proxy-chain/pull/157/commits/f04773e287a9ac2105375f389694d2bde596849b)

src/server.js
mnmkng: Yeah, you're right. It stopped working from 0.4.9 to 1.0.0.

src/utils/valid_headers_only.js
B4nan: that is a separate commit, not separate PR. our point is that you do many changes in a single PR, and moving files around basically disables diffing, we can't easily tell whether you just moved them or also changed something in the implementation. 

same as with the linting problems

szmarczak:
> ignore invalid ...

@mnmkng This shouldn't be breaking as the upstream proxy should respond with 401.

src/anonymize_proxy.js
mnmkng: I'm talking about the switch from `parseUrl` to `new URL` without all the additional props and defaults `parseUrl` added. I'm pretty sure that given the number of use cases we have, there will be some people with some edge-case proxy URLs that will start failing for them.

src/tools.js
szmarczak: I was initially thinking about removing this from `proxy-chain`, but this would be too breaking as it's used in too many places. Maybe someday in the future.

package.json
```diff
@@ -54,6 +54,7 @@
     "eslint-plugin-react": "^7.0.1",
     "express": "^4.16.2",
     "faye-websocket": "^0.11.3",
+    "got-scraping": "^3.2.4-beta.0",
```
mnmkng, line 57: Btw, why got-scraping? I can't find it anywhere else in the changes.

szmarczak: test/server.js lines 487-551

> why got-scraping?

`request` proxy implementation fails to send proper IPv6 target sends `...` instead of `[...]`

src/server.js
mnmkng: Ok. I'm just pointing it out because previously we expected a port number here, so if we don't anymore, somewhere down the line we have to be sure that `''` will also work.

src/anonymize_proxy.js
szmarczak: > there will be some people with some edge-case proxy URLs that will start failing for them.

Then they will need to switch to WHATWG URL (which is a good change). For `proxy-chain` there's no need to have this. This rewrite will be a major anyway.

src/utils/valid_headers_only.js
mnmkng: Yup, what @B4nan says. I have to read the whole functions to see if they're the same or not. And even then I'm not really sure because the code is not side-by-side. It's a quality-of-reviewer's-life suggestion :)

szmarczak: I agree, though it's not much noise given those are separate test files. If a PR is big, it's good to review it on a per-commit basis.

package.json
mnmkng: Ah, thanks. The diff was collapsed so cmd+F failed me :)

szmarczak: Hint: press <kbd>.</kbd> :D

szmarczak: Although the search (sidebar) feature is quite broken, instead of searching all files it just searches through the current file :( So need to browser file by file.

src/anonymize_proxy.js
mnmkng: Yeah, they will. But I want to make sure we don't forget about this and do not merge `proxy-chain` v2 into SDK v2. We can only add it to v3 because there's a lot of under the hood changes.

Thinking about it more. I'm not quite sure how this will work for all the proxy URLs that are used from the Apify Console UI. Those people won't easily change to a WHATWG URL and there might be hell on support because of this.

szmarczak: > I'm not quite sure how this will work for all the proxy URLs that are used from the Apify Console UI.

I don't understand. What would change? Their actors will work exactly the same without any changes. The old `parseUrl` wraps WHATWG URL already. They can't tinker with server-side code, can they?

If you mean providing their own proxy URLs, the behavior remains unchanged.

mnmkng: I guess we'll have to see :)

mnmkng:
> This shouldn't be breaking as the upstream proxy should respond with 401.

If that's the case, I think we should keep the locally thrown validation errors because I don't trust the upstream proxies to provide useful error messages.

`...cannot contain colon...` is much better than `401 Unauthorized`

szmarczak:
> If that's the case, I think we should keep the locally thrown validation errors because I don't trust the upstream proxies to provide useful error messages.

An incorrect colon in username is the same error as typing `a` instead of `e` (`0` vs `O`, `l` vs `I` vs `1`) or vice versa.
In reality there's no difference if `cannot contain colon` is thrown instead of `401`.
Invalid credentials? Invalid. Why can't we move on.

I'll implement this anyway.

szmarczak:
@mnmkng @B4nan ptal

mnmkng (approved):

B4nan (approved):

PR #158, opened 2021-09-16
wip: simplify tunnel

szmarczak:
#### this PR

- [x] simplify `TcpTunnel` https://blog.apify.com/tunneling-arbitrary-protocols-over-http-proxy-with-static-ip-address-b3a2222191ff/
- [x] turn the 5 parameters of `direct` and `chain` into object

#### [wip: simplify forward](https://github.com/apify/proxy-chain/pull/153)

- [x] remove babel
- [x] added https support
  - [ ] test
- [x] require node.js 14

#### [wip: simplify custom response](https://github.com/apify/proxy-chain/pull/154)

- [x] simplify custom response
- [x] use shared eslint config

#### [wip: simplify direct](https://github.com/apify/proxy-chain/pull/155)

- [x] simplify direct

#### [wip: simplify chain](https://github.com/apify/proxy-chain/pull/156)

- [x] simplify chain
  - [x] remove handler base

#### [wip: simplify tools](https://github.com/apify/proxy-chain/pull/157)

- [x] simplify tools
  - [x] remove `parseUrl(...)` in favor of `new URL(...)`
  - [x] remove `parseHostHeader(...)` in favor of `new URL(...)`
  - [x] remove `isInvalidHeader(...)` (unused)
  - [x] tests: remove buggy `findFreePort` in favor of `port: 0`
  - [x] simplify `redactUrl`
  - [x] remove `addHeader` (leftover from #156)
- [x] add IPv6 tests

There's a slight behavior change in `parseAuthorizationHeader`. It now returns `{ type: '', data: '' }` (instead of `null`) if the `proxy-authorization` header is invalid.

#### TODO (those will be separate PRs)

- [ ] simplify server
- [ ] use typescript
- [ ] more descriptive errors
- [ ] enable coverage


src/chain.js
```diff
@@ -2,14 +2,10 @@ const http = require('http');
 const { getBasic } = require('./utils/get_basic');
 
 /**
- * @param {http.ClientRequest} request
- * @param {net.Socket} source
- * @param {buffer.Buffer} head
- * @param {*} handlerOpts
- * @param {*} server
+ * @param {*} options
```
B4nan, line 5: would be good to document each of the options

B4nan (approved):
codewise looking good, let's see what others think about the implementation

src/chain.js
szmarczak: They will be removed when we switch to TypeScript. Won't have any effect then.

B4nan: They would be converted to an interface, not removed, right?

szmarczak: yep, but the jsdocs won't be there since TypeScript will make them moot

mnmkng: Can we JSDoc an interface? Aka have the comments explaining what's what on the interface itself? I mean, I have literally no idea what some of those are.

src/server.js
```diff
@@ -169,13 +169,15 @@ class Server extends EventEmitter {
                 handlerOpts = result;
                 handlerOpts.srcHead = head;
 
+                const data = { request, source: socket, head, handlerOpts, server: this };
```
mnmkng, line 172: Would it make sense to make a class out of this instead of a plain object? It seems quite complex and we will be making an interface anyway. Just sayin'. 🤷

src/chain.js
B4nan: sure, that is what i'd like to see in the end. we don't need to comment every single property, but some are definitely worth a few extra words.

```ts
interface FooBar {
  /** this is a foo */
  foo: number;
  /** this is a bar */
  bar: string;
}
```

those comments won't be presented via intellisense, but they are there when hovering over the propert

image

src/tcp_tunnel_tools.js
```diff
@@ -1,78 +1,66 @@
...
+        if (verbose) console.log(...args);
     };
 
+    server.log = log;
+
     server.on('connection', (srcSocket) => {
-        const { port } = server.address();
+        runningServers[address].connections.push(srcSocket);
 
-        runningServers[port].connections.push(srcSocket);
```
mnmkng, line 33: I see that we're now using an address that changes betwen IPv4 and 6, but now the use of a closured variable, which is initialized as `undefined` and only set in some other callback 20 lines later is not very readable. Is there a way to improve it?

src/tcp_tunnel_tools.js
```diff
@@ -1,78 +1,66 @@
...
         });
+    });
 
-        tunnel.run();
+    const promise = new Promise((resolve, reject) => {
+        server.once('error', reject);
 
-        srcSocket.on('data', onConnData);
-        srcSocket.on('close', onConnClose);
-        srcSocket.on('error', onConnError);
```
mnmkng, line 53: Are we removing all the logs? Why?

src/tcp_tunnel_tools.js
```diff
@@ -81,26 +69,19 @@ function createTunnel(proxyUrl, targetHost, providedOptions = {}, callback) {
...
-    })
-        .then((serverExists) => new Promise((resolve) => {
-            if (!serverExists) return resolve(false);
-            runningServers[port].server.close(() => {
-                delete runningServers[port];
-                resolve(true);
-            });
-        }));
+function closeTunnel(address, closeConnections) {
+    if (!runningServers[address]) return false;
```
mnmkng, line 73: I would make a variable out of `runningServers[address]`. It's used in 3 places and `runningServers[address].connections` and `runningServers[address].server.close()` is a bit too much.

src/tcp_tunnel_tools.js
```diff
@@ -81,26 +69,19 @@ function createTunnel(proxyUrl, targetHost, providedOptions = {}, callback) {
...
 
-    return nodeify(promise, callback);
+    if (closeConnections) {
+        for (const connection of runningServers[address].connections) {
+            connection.destroy();
+        }
+    }
+
+    runningServers[address].server.close();
```
mnmkng, line 81: Why ignore the close callback?

src/tcp_tunnel_tools.js
```diff
@@ -81,26 +69,19 @@ function createTunnel(proxyUrl, targetHost, providedOptions = {}, callback) {
 
 module.exports.createTunnel = createTunnel;
 
-function closeTunnel(serverPath, closeConnections, callback) {
-    const [hostname, port] = serverPath.split(':');
-    if (!hostname) throw new Error('serverPath must contain hostname');
-    if (!port) throw new Error('serverPath must contain port');
```
mnmkng, line 87: I'm not saying this validation is great, but `closeTunnel` is public API. So we should validate this somehow or document the shape of `address` properly and enforce its type.

src/chain.js
```diff
@@ -2,14 +2,10 @@ const http = require('http');
...
  */
-const chain = (request, source, head, handlerOpts, server) => {
-    if (head.length > 0) {
+const chain = ({ request, source, head, handlerOpts, server, plain }) => {
```
jancurn, line 7: Tiny thing: If `plain` is a boolean, in the Apify codebase by convention we prefix those variables with `has` or `is`. E.g. `isPlain`

src/tcp_tunnel_tools.js
```diff
@@ -81,26 +69,19 @@ function createTunnel(proxyUrl, targetHost, providedOptions = {}, callback) {
...
-
-    const promise = new Promise((resolve) => {
```
mnmkng, line 89: I read in the docs that this returns a promise if the callback is not provided. I wonder if it makes sense to remove that capability and potentially break someone's code. What's the benefit?

src/chain.js
jancurn: Indeed, without the comments, it will be very hard for anyone to maintain this code in the future

mnmkng (changes requested):

src/chain.js
szmarczak: Good idea.

src/server.js
szmarczak: Nope. Why a class?

src/tcp_tunnel_tools.js
szmarczak: Ah, sorry, forgot about them by accident. Will restore this.

src/tcp_tunnel_tools.js
szmarczak: We could make a function `getAddress`. Wdyt?

src/tcp_tunnel_tools.js
szmarczak: > So we should validate this somehow

This return `false` if the server does not exist. If anybody wants to reassure an actual server was closed they should make an assertion.

src/server.js
mnmkng: To encapsulate some logic which seems to be used in a lot of places and give it a name. Readability and documentation purposes. It's not important though. Just something that occured to me when I saw a plain object stuffed with 5 properties.

src/tcp_tunnel_tools.js
mnmkng: Yeah, sounds good.

src/tcp_tunnel_tools.js
szmarczak: > I wonder if it makes sense to remove that capability and potentially break someone's code.

Indeed, this is breaking.

> What's the benefit?

There's no need for this code to be asynchronous. The `error` argument is either undefined or an Error if the server was already closed, which by the current usage here, is never the case.

src/tcp_tunnel_tools.js
szmarczak: https://github.com/apify/proxy-chain/pull/158#discussion_r710264562

src/server.js
szmarczak: Hmm... The documentation purposes is very convincing 🤔 I don't think it's worth it for 5 properties though. TypeScript interface should be sufficient to be honest.

src/tcp_tunnel_tools.js
mnmkng: It's not about making sure that a server was closed, but knowing that you passed in the correct argument.

src/tcp_tunnel_tools.js
mnmkng: Yeah, it's true that the error argument is rather pointless. But someone might want to just plainly wait for the server to close before doing something else. Now they can't.

src/chain.js
szmarczak: @jancurn With due respect, the previous implementation was very hard to maintain in contrast to this one. TypeScript will make it more understandable as well. By knowing the Node.js API and reading the code, the meaning of the properties is very clear. I'm not saying that we mustn't doc this. Just doing so won't help that much IMHO. Maybe this is just me and needs to be documented 100%.

szmarczak: I have had a discussion with @mnmkng and decided to add comments anyway. Here's the golden rule I very liked:

> Always leave the campground cleaner than you found it

:D

src/tcp_tunnel_tools.js
szmarczak: > It's not about making sure that a server was closed, but knowing that you passed in the correct argument.

I don't understand how relevant this is. The purpose of this function is to close a server.
Can you explain the reasoning please?

src/tcp_tunnel_tools.js
szmarczak: `error` is logged inside `chain`. `data` makes the logs too messy. I'll just log on `close`.

szmarczak: Fixed.

src/direct.js
```diff
@@ -1,23 +1,30 @@
 const net = require('net');
 
 /**
- * @param {http.ClientRequest} request
- * @param {net.Socket} source
- * @param {buffer.Buffer} head
- * @param {*} handlerOpts
- * @param {*} server
+ * @typedef Options
```
szmarczak, line 4: Currently this is duplicated but will be a single interface in the TypeScript PR.

src/tcp_tunnel_tools.js
mnmkng: When I pass in an invalid address, it tells me that the server is closed, instead of telling me that the address is invalid. The previous validation made sure that you're at least passing in something reasonable.

szmarczak: > When I pass in an invalid address, **it tells me that the server is closed**, instead of telling me that the address is invalid.

With the previous implementation on `master`, it returns `false` as well if you provide a valid address but [the server may haven't existed at all](https://github.com/apify/proxy-chain/blob/0043ac9db2c04fffdd53fe7ee7d3a11183648555/src/tcp_tunnel_tools.js#L94). Following your thinking, it should throw an error because the server can't be closed. And to check if the server was already closed, we would have to store it, which would do a memory leak. So the solution would be to expose it, so we don't have to care about storing it.

You get it wrong. `false` doesn't tell you that the server is closed, it says that it failed to do so.

As I've [already mentioned](https://github.com/apify/proxy-chain/pull/158#discussion_r710264562), an `error` is passed to the callback only and only if the server was already closed, which is never the case here.

> The previous validation made sure that you're at least passing in something reasonable.

How does this explain that it should throw? We want to close a server, not to check if the server actually exists. If the user code assumes the server exists, they should assert that this returns `true` - and this already validates the argument.

---

I propose to return the actual server and get rid of `closeTunnel`.

mnmkng (approved):

PR #161, opened 2021-09-27
wip: move to TypeScript

szmarczak:


src/chain.ts
```diff
@@ -14,47 +17,60 @@ const createHttpResponse = (statusCode, message) => {
...
+        head,
+        handlerOpts,
+        server,
+        isPlain,
+    }: {
+        request: { url?: string },
+        sourceSocket: net.Socket,
+        head?: Buffer,
+        handlerOpts: HandlerOpts,
+        server: EventEmitter & { log: (...args: any[]) => void; },
```
B4nan, line 43: `unknown` might be better than `any` here

src/forward.ts
```diff
@@ -39,13 +49,13 @@ const forward = async (request, response, handlerOpts) => new Promise(async (res
...
     }
 
-    const fn = origin.startsWith('https:') ? https.request : http.request;
+    const fn = origin!.startsWith('https:') ? https.request : http.request;
 
-    const client = fn(origin, options, async (clientResponse) => {
+    const client = fn(origin!, options as unknown as http.ClientRequestArgs, async (clientResponse) => {
         try {
             // This is necessary to prevent Node.js throwing an error
             let { statusCode } = clientResponse;
```
B4nan, line 57: we could handle the optionallity right here to get rid of all the `!` in following lines. either via casting `clientResponse` or by changing this to `let statusCode = clientResponse.statusCode!`

(personally i am not in the "prefer object destructing everywhere blindly" :P)

src/request_error.ts
```diff
@@ -6,8 +6,12 @@
...
  * and for the 407 status the Proxy-Authenticate header will be added.
  */
-class RequestError extends Error {
-    constructor(message, statusCode, headers) {
+export class RequestError extends Error {
+    statusCode: number;
+
+    headers?: Record<string, string>;
+
+    constructor(message: string, statusCode: number, headers?: Record<string, string>) {
```
B4nan, line 14: we can define the new properties directly in constructor

src/server.ts
```diff
@@ -31,12 +33,67 @@ const { handleCustomResponse } = require('./custom_response');
 const DEFAULT_AUTH_REALM = 'ProxyChain';
 const DEFAULT_PROXY_SERVER_PORT = 8000;
 
+type ConnectionStats = {
+    srcTxBytes: number;
+    srcRxBytes: number;
+    trgTxBytes: number | null;
```
B4nan, line 39: maybe we should mark the nullable properties as optional too?

src/server.ts
```diff
@@ -95,15 +157,15 @@ class Server extends EventEmitter {
...
         if (this.verbose) {
             const logPrefix = connectionId ? `${connectionId} | ` : '';
             console.log(`ProxyServer[${this.port}]: ${logPrefix}${str}`);
         }
     }
 
-    onClientError(err, socket) {
-        this.log(socket.proxyChainId, `onClientError: ${err}`);
+    onClientError(err: NodeJS.ErrnoException, socket: net.Socket): void {
+        this.log((socket as any).proxyChainId, `onClientError: ${err}`);
```
B4nan, line 168: maybe we should use own extension of `net.Socket` (interface), this type casting because of `proxyChainId` is quite common

src/server.ts
```diff
@@ -31,12 +33,67 @@ const { handleCustomResponse } = require('./custom_response');
...
+export class Server extends EventEmitter {
+    port: number;
+
+    prepareRequestFunction?: PrepareRequestFunction;
+
+    authRealm: unknown;
+
+    verbose: boolean;
+
+    server: http.Server | null;
```
B4nan, line 89: looks like we always have the server set (done in constructor), why the `| null`? there are many `!` later in the code because of it

B4nan (changes requested):
looking good, but i believe we could get rid of many type casts and non-null assertions

jest.config.ts
```diff
@@ -0,0 +1,21 @@
+import type { Config } from '@jest/types';
+
+export default async (): Promise<Config.InitialOptions> => ({
```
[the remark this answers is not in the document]
szmarczak: I just copied this from other repo, will fix this :+1:

.eslintrc.json
```diff
@@ -1,3 +1,18 @@
...
+    "extends": [
+      "@apify/ts"
+    ],
+    "parserOptions": {
+      "project": "tsconfig.eslint.json"
+    },
+    "overrides": [
+      {
+        "files": [
+          "test/**/*.js"
```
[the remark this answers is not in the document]
szmarczak: The tests are very JS-ly complicated and I believe it would take more time than the source code itself.

src/server.ts
szmarczak: No, it's either number or null. It's never undefined. I was thinking about moving to undefined but that would be breaking.

src/server.ts
szmarczak: That's because initially it isn't created I think

B4nan: wdym? the instance [gets created in constructor](https://github.com/apify/proxy-chain/pull/161/files/b326729704a75725b2333305289c0ca2e9ce8571#diff-8a8ae07582c9d433ec8c2e5c4310ff8901e604f4965c5b90a49117ad46c47595R145), so it should be there unless you create the instance via `Object.create()`

.eslintrc.json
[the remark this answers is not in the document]
B4nan: the motivation for me would be to run tests directly on source code without compilation - i believe that should be possible even if we keep the tests in JS and use ts-node on them (and importing stuff from `src` instead of `dist`)

wdyt about that? definitely not a priority to rewrite it if it would be too time consuming

src/server.ts
szmarczak: Indeed it is, not sure what I was thinking, I'll fix this

.eslintrc.json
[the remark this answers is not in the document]
szmarczak: That sounds very cool, I'll try this :)

src/server.ts
szmarczak: Actually it's set to null in the `close` function.

src/chain.ts
szmarczak: It has to be any, otherwise `server.ts` errors that `unknown` cannot be cast to `string` etc.

src/chain.ts
```diff
@@ -14,47 +17,60 @@ const createHttpResponse = (statusCode, message) => {
...
+            request.url!,
         ],
     };
 
     if (proxy.username || proxy.password) {
         options.headers.push('proxy-authorization', getBasic(proxy));
     }
 
-    const client = http.request(proxy.origin, options);
+    const client = http.request(proxy.origin, options as unknown as http.ClientRequestArgs);
```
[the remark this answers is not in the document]
szmarczak: The issue is that `@types/node` doesn't support an array for `headers`. We're using an array because it's easier to mutate it and it's much faster than a plain object.

src/anonymize_proxy.ts
```diff
@@ -1,18 +1,18 @@
...
 /**
  * Parses and validates a HTTP proxy URL. If the proxy requires authentication, then the function
  * starts an open local proxy server that forwards to the upstream proxy.
- * @param proxyUrl
- * @param callback Optional callback that receives the anonymous proxy URL
- * @return If no callback was supplied, returns a promise that resolves to a String with
- * anonymous proxy URL or the original URL if it was already anonymous.
  */
-const anonymizeProxy = (proxyUrl, callback) => {
+export const anonymizeProxy = (proxyUrl: string, callback?: (error: Error | null) => void): Promise<string> => {
```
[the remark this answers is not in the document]
szmarczak: What comment are you referring to? This always returns a Promise.

> maybe we could move the nodeify call to the callsite instead of the method definition?

wdym?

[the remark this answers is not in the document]
B4nan: This one: 

> @return If no callback was supplied, returns a promise that resolves to a String with
> * anonymous proxy URL or the original URL if it was already anonymous.

Which should be caused by the `nodeify` call that turns the promise function to a function with callback, and if the callback is provided, it won't return the value, but feed it into the callback argument instead.

Maybe I got it wrong, but if that is the case, I though we could move the `nodeify` call from inside this method to the call site, so to where the `anonymizeProxy` is used (and where the callback style signature might be required).

.eslintrc.json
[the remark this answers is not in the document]
szmarczak: Fixed! :D

src/anonymize_proxy.ts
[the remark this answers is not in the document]
szmarczak: > Which should be caused by the nodeify call that turns the promise function to a function with callback, and if the callback is provided, it won't return the value, but feed it into the callback argument instead.

Well, in the previous code it was always returning a Promise as well 🤔 Should we change this?

> Maybe I got it wrong, but if that is the case, I though we could move the nodeify call from inside this method to the call site, so to where the anonymizeProxy is used (and where the callback style signature might be required).

Well the issue is that it's used outside this package as well.

src/anonymize_proxy.ts
```diff
@@ -53,19 +53,16 @@ const anonymizeProxy = (proxyUrl, callback) => {
...
 };
 
-module.exports.anonymizeProxy = anonymizeProxy;
-
 /**
  * Closes anonymous proxy previously started by `anonymizeProxy()`.
  * If proxy was not found or was already closed, the function has no effect
  * and its result if `false`. Otherwise the result is `true`.
- * @param anonymizedProxyUrl
- * @param closeConnections If true, pending proxy connections are forcibly closed.
```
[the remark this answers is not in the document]
B4nan: The other part that describes how it works when the flag is set to false (or not set at all), that part seem to have some value for me. 

> If `false`, the function will wait until all connections are closed, which can take a long time.

szmarczak:
GitHub Actions have stopped for this PR and I have no idea why 🤔

src/server.ts
B4nan: I see, still it makes more sense to me to have it defined as mandatory, if you call `close` it feels like you should not use any of the methods anyway.

We are talking about how to type this, the functionality itself shows that we don't check if the property is set, hence it feels like it should be set to mandatory and we could escape the strict type only in the `close` method instead of having `!` almost everywhere.

B4nan:
Looks like GH actions are broken

image

https://www.githubstatus.com/

src/server.ts
szmarczak: So we can do this in two ways. Should I make it `http.Server` or `const server = this.server!` ?

B4nan: I'd probably rather type it as mandatory instead of introducing local variable with `!` (especially if there are more than one methods that would have to do it). But both are fine.

src/chain.ts
```diff
@@ -50,7 +51,7 @@ export const chain = (
         throw new Error(`Unexpected data on CONNECT: ${head.length} bytes`);
     }
 
-    const { proxyChainId } = sourceSocket as unknown as { proxyChainId: unknown };
+    const { proxyChainId } = sourceSocket as Socket;
```
B4nan, line 54: what about using the type directly in `ChainOpts`? we might need to make the `id` optional to allow passing raw `net.Socket` instance, but that seems totally fine

src/direct.ts
```diff
@@ -75,7 +76,7 @@ export const direct = (
         }
     });
 
-    const { proxyChainId } = sourceSocket as unknown as { proxyChainId: unknown };
+    const { proxyChainId } = sourceSocket as Socket;
```
B4nan, line 79: same here with `DirectOpts`

src/anonymize_proxy.ts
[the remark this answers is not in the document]
B4nan: > Well the issue is that it's used outside this package as well.

I see, ok then let's keep it as it is.

src/request_error.ts
```diff
@@ -6,17 +6,16 @@
...
+export class RequestError extends Error {
+    constructor(
+        message: string,
+        public statusCode: number,
+        public headers?: Record<string, string>,
+    ) {
         super(message);
         this.name = RequestError.name;
-        this.statusCode = statusCode;
         this.headers = headers;
```
B4nan, line 17: this line seems obsolote now, as it will be assigned directly from ctor parameter

src/server.ts
```diff
@@ -197,12 +261,12 @@ class Server extends EventEmitter {
...
      * @param head The first packet of the tunneling stream (may be empty)
      */
-    async onConnect(request, socket, head) {
+    async onConnect(request: http.IncomingMessage, socket: Socket, head: Buffer): Promise<void> {
         try {
             const handlerOpts = await this.prepareRequestHandling(request);
             handlerOpts.srcHead = head;
 
-            const data = { request, sourceSocket: socket, head, handlerOpts, server: this };
+            const data = { request, sourceSocket: socket, head, handlerOpts: handlerOpts as ChainOpts, server: this, isPlain: false };
```
B4nan, line 269: we could move the type assertion to `handlerOpts` definition (line 266)

B4nan (approved):
looking good, just two small nitpicks

src/server.ts
szmarczak: We actually can't do this because it's different in case of `direct` handler.

PR #175, opened 2021-11-02
Add support for custom dns lookup

thomas510111:
Default dns.lookup() is running on libuv threadpool and on high load or issues with the dns server it can cause increased latency in getting the initial connection.
https://github.com/nodejs/node/issues/8436

In my opinion the best way to "patch" this is allow users to use their own dns lookup function.


thomas510111:
By the way, I havent worked with typescript and couldnt figure out how to properly declare type of dnsLookup.

src/chain.ts
```diff
@@ -22,11 +22,13 @@ interface Options {
     headers: string[];
     path?: string;
     localAddress?: string;
+    lookup?: any;
```
szmarczak, line 25: Please type this properly.

szmarczak (changes requested):

src/chain.ts
szmarczak: I think it should be possible to `dns['lookup']` or `typeof dns['lookup']`

freeway7:
How to set custom DNS server?

PR #176, opened 2021-11-02
feat: `closeConnection` by id

thomas510111:


src/server.ts
```diff
@@ -564,6 +564,20 @@ export class Server extends EventEmitter {
         return result;
     }
 
+    /**
+     * Forcibly close a specific pending proxy connection.
+     */
+    closeConnection(connectionId: unknown): void {
```
szmarczak, line 570: Why would you need this?

janisblaus: because there are no timeouts and if the remote server keeps connection open/hangs up then it will keep the connection open forever, I had the same issue. I resolved it similarly as in this PR but checking if remote upstream server hasn't sent any bytes yet so its considered as a timeout

szmarczak: Sorry for late reply. I think you could override https://github.com/apify/proxy-chain/blob/a6d51fb526258c82308ac98215afd5349330f9dd/src/server.ts#L203 (and call `super.onConnection(socket)`).

Alternatively, you could access `proxyChainServer.server` to get the underlying native `http` server.

thomas510111: We mainly need that to be able to terminate specific connections, if they overuse their data plans.

szmarczak: Makes sense. Thanks for explanation :)

src/server.ts
```diff
@@ -564,6 +564,20 @@ export class Server extends EventEmitter {
...
+        this.log(null, 'Closing pending socket');
+
+        const socket = this.connections.get(connectionId);
+        if (!socket) return;
+
+        socket.destroy();
```
szmarczak, line 576: Destroy actually doesn't destroy the socket (however it tries to close it). See https://github.com/nodejs/node/issues/27428

PR #285, opened 2022-08-18
feat: custom connect http server

szmarczak:
This would enable to use CONNECT for SERP requests. Currently using [`hpagent`](https://github.com/delvedor/hpagent) doesn't work when doing SERPs, because it only supports CONNECT proxying.

/cc @AndreyBykov

### DIFF
--- a/src/custom_connect.ts
+++ b/src/custom_connect.ts
@@ -12,7 +12,7 @@ export const customConnect = async (socket: net.Socket, server: http.Server): Pr
     server.emit('connection', socket);
 
     return new Promise((resolve) => {
-        if (socket.closed) {
+        if (socket.destroyed) {
             resolve();
             return;
         }
--- a/src/custom_connect.ts
+++ b/src/custom_connect.ts
@@ -1,0 +1,24 @@
+import net from 'net';
+import type http from 'http';
+import { promisify } from 'util';
+// import { countTargetBytes } from './utils/count_target_bytes';
+
+const asyncWrite = promisify(net.Socket.prototype.write);
+
+export const customConnect = async (socket: net.Socket, server: http.Server): Promise<void> => {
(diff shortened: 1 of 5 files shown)


src/custom_connect.ts
```diff
@@ -0,0 +1,24 @@
+import net from 'net';
+import type http from 'http';
+import { promisify } from 'util';
+// import { countTargetBytes } from './utils/count_target_bytes';
+
+const asyncWrite = promisify(net.Socket.prototype.write);
+
+export const customConnect = async (socket: net.Socket, server: http.Server): Promise<void> => {
+    // countTargetBytes(socket, socket);
```
szmarczak, line 9: Do we want to count bytes here? Also both `proxy-chain` v1 and v2 we don't count bytes when using `customResponseHandler`. How are users accounted for data transferred when using SERP?

mnmkng: SERPs are billed per request, not per GBs

szmarczak: Ah ok. Do we want to track bytes transferred anyway?

jancurn (commented):
Please can you add info to README and explain there where/how this feature is useful?

src/custom_connect.ts
jancurn: If bytes are simple to count, I'd add it. If it's complicated, fuk it.

jirimoravcik (approved):

fnesveda (approved):

jancurn (approved):
Please can you add info about this to README?

szmarczak:
Will do, multitasking 😅

szmarczak:
@jancurn [Added docs](f63a0e8fe7b812e9cbc13ad4120dd826d780d836) ptal.

PR #316, opened 2022-09-28
feat: custom bad gateway statuses

szmarczak:


src/chain.ts
```diff
@@ -118,7 +119,10 @@ export const chain = (
             if (isPlain) {
                 sourceSocket.end();
             } else {
-                sourceSocket.end(createHttpResponse(502, ''));
+                const unauthorized = [401, /* 403, */ 407];
+                const status = unauthorized.includes(response.statusCode!) ? 597 : 590;
```
jancurn, line 123: It would be better and less error-prone to put these numbers in a constant

src/statuses.ts
```diff
@@ -0,0 +1,7 @@
+export const statuses: {[code: string]: number | undefined} = {
```
jancurn, line 1: Comment?

jancurn (approved):
Neat idea, just a few comments

src/chain.ts
szmarczak: Doesn't matter but I guess it's more readable, will do that 👍

szmarczak: Doesn't matter but I guess it's more readable, will do that 👍

src/statuses.ts
szmarczak: What needs to be commented here?

jancurn: just what is this lookup table for

src/chain.ts
jancurn: I mean to have all the new codes with a special meaning in a constant dictionary, just to make it clear what they mean, without having to look into README

src/statuses.ts
szmarczak: I guess I can just rename `code` to `errorCode` and alias `number` as `HttpStatusCode`?

jirimoravcik (approved):

szmarczak:
@jancurn Added comments for https://github.com/apify/proxy-chain/blob/custom-bad-gateway/src/statuses.ts PTAL :)

jancurn (approved):

jancurn:
Great, thanks @szmarczak

Issue #362, opened 2022-12-06
A lot of TCP connections


Rumatoid:
proxy-chain generate a lot of TCP connections
without PC I have about 50 connections, but 150 with PC

Im using ProxyChain.Server with prepareRequestFunction

kami4ka:
Up! I Found the same issue. Connections are not closing.

jirimoravcik:
Hello guys,
thanks for reporting this issue. You have any minimal code to reproduce the problem?
Thanks

Rumatoid:
> Hello guys, thanks for reporting this issue. You have any minimal code to reproduce the problem? Thanks

Just use prepareRequestFunction and watch for TCP connections (in Linux for example)

kami4ka:
I suppose that the reason is that `timeout` event from the socket is not handled, so the connection wouldn't be closed, as in case of timeout it should be closed manually.

jirimoravcik:
Hello guys,
thanks for your effort to report this issue.
We have released version 2.2.1 that should not leak TCP connections anymore.

PR #407, opened 2023-02-16
fix: Handle socket timeout to prevent connections from leaking

jirimoravcik:
https://nodejs.org/api/net.html#event-timeout `Emitted if the socket times out from inactivity. This is only to notify that the socket has been idle. The user must manually close the connection.`

### DIFF
--- a/src/server.ts
+++ b/src/server.ts
@@ -203,6 +203,11 @@ export class Server extends EventEmitter {
 
             this.connections.delete(unique);
         });
+        // We have to manually destroy the socket if it timeouts.
+        // This will prevent connections from leaking and close them properly.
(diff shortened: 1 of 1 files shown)


fnesveda (approved):
Nice! How did you find this?

jirimoravcik:
> Nice! How did you find this?
`netstat -pnat | grep 8000`
https://serverfault.com/questions/738300/why-are-connections-in-fin-wait2-state-not-closed-by-the-linux-kernel

kami4ka:
I thought it was my finding https://github.com/apify/proxy-chain/issues/362#issuecomment-1356260697 😂
Nice job, guys!

PR #411, opened 2023-02-27
feat: Add `metadata` parameter to the tunnelConnectResponded event

jirimoravcik:
Also add tunnelConnectError event.

This is used to pass information that was passed by `prepareRequestFunction` back via events

### DIFF
--- a/src/server.ts
+++ b/src/server.ts
@@ -52,6 +52,7 @@ type HandlerOpts = {
     localAddress?: string;
     ipFamily?: number;
     dnsLookup?: typeof dns['lookup'];
+    metadata?: Record<string, unknown>;
 };
 
 export type PrepareRequestFunctionOpts = {
@@ -73,6 +74,7 @@ export type PrepareRequestFunctionResult = {
     localAddress?: string;
     ipFamily?: number;
     dnsLookup?: typeof dns['lookup'];
+    metadata?: Record<string, unknown>;
 };
 
 type Promisable<T> = T | Promise<T>;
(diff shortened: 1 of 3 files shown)


jancurn (approved):

src/server.ts
```diff
@@ -73,6 +74,7 @@ export type PrepareRequestFunctionResult = {
     localAddress?: string;
     ipFamily?: number;
     dnsLookup?: typeof dns['lookup'];
+    metadata?: Record<string, unknown>;
```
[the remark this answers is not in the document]
jancurn: btw does this need to be a dictionary? why not any object user-supplied object?

fnesveda (approved):
Agreed with Honza's points, plus it would be nice to have a test for these events, unless it would be too much work to add.

Otherwise it looks good 🙂

README.md
```diff
@@ -72,6 +72,10 @@ const server = new ProxyChain.Server({
             // If "requestAuthentication" is true, you can use the following property
             // to define a custom error message to return to the client instead of the default "Proxy credentials required"
             failMsg: 'Bad username or password, please try again.',
+
+            // Optional metadata that will be passed back via
+            // `tunnelConnectResponded` or `tunnelConnectFailed` events
+            metadata: { userId: '123' },
```
jancurn, line 78: sorry, just one last note 🙈 wouldn't it make more sense to call this `tag` or `customTag` rather than `metadata`?  metadata has a specific meaning. Your call @jirimoravcik

jirimoravcik: Good point, I went with `customTag`, smaller chance of a meaning conflict due to the name

Issue #521, opened 2023-08-19
Add suport for Bun runtime - Issue with `net.Socket.write`


JuroOravec:
Hi, I have a web scraper built on top of the [Crawlee framework](https://github.com/apify/crawlee). I wanted to run it with [Bun](https://bun.sh/) instead of Node. However, it failed, and the stack trace (at the bottom) led to `proxy-chain`.

I don't think the issue is with Bun. I tried running the following script with Bun, and both `new Socket()` and `socket.write` were defined:

```js
const { Socket } = require('net');
console.log({
  socket: new Socket(),
  write: (new Socket()).write,
});
```

Error stack trace:

```sh
2 | Object.defineProperty(exports, "__esModule", { value: true });
3 | exports.customConnect = void 0;
4 | const tslib_1 = require("tslib");
5 | const net_1 = tslib_1.__importDefault(require("net"));
6 | const util_1 = require("util");
7 | const asyncWrite = (0, util_1.promisify)(net_1.default.Socket.prototype.write);
                                            ^
TypeError: undefined is not an object (evaluating 'net_1.default.Socket.prototype.write')
      at /Users/presenter/repos/proj-name/node_modules/proxy-chain/dist/custom_connect.js:7:41
      at globalThis (/Users/presenter/repos/proj-name/node_modules/proxy-chain/dist/custom_connect.js:26:24)
      at require (:1:20)
      at /Users/presenter/repos/proj-name/node_modules/proxy-chain/dist/server.js:21:6
      at globalThis (/Users/presenter/repos/proj-name/node_modules/proxy-chain/dist/server.js:546:17)
      at require (:1:20)
      at /Users/presenter/repos/proj-name/node_modules/proxy-chain/dist/index.js:5:8
      at globalThis (/Users/presenter/repos/proj-name/node_modules/proxy-chain/dist/index.js:8:59)
      at require (:1:20)
      at /Users/presenter/repos/proj-name/node_modules/@crawlee/browser-pool/anonymize-proxy.js:4:6
      at globalThis (/Users/presenter/repos/proj-name/node_modules/@crawlee/browser-pool/anonymize-proxy.js:33:30)
      at require (:1:20)
      at /Users/presenter/repos/proj-name/node_modules/@crawlee/browser-pool/puppeteer/puppeteer-controller.js:7:6
      at globalThis (/Users/presenter/repos/proj-name/node_modules/@crawlee/browser-pool/puppeteer/puppeteer-controller.js:111:30)
[...]

PR #522, opened 2023-08-20
refactor: Avoid accessing prototype in custom_connect

JuroOravec:
See the discussion here: https://github.com/apify/proxy-chain/issues/521#issuecomment-1685378805

NOTE: I tried running `npm run test` to ensure everything works. But the tests were failing because of other error (below) even before I made the change, so I can't ensure integrity.

```sh
$ npm run test

> proxy-chain@2.3.0 test
> nyc cross-env NODE_OPTIONS=--insecure-http-parser mocha --bail



(node:23123) Warning: Using insecure HTTP parsing
(Use `node --trace-warnings ...` to show where the warning was created)
  1) supports localAddress

  0 passing (2s)
  1 failing

  1) supports localAddress:
     Error: Timeout of 2000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (/Users/presenter/repos/proxy-chain/test/server.js)
      at listOnTimeout (node:internal/timers:557:17)
      at processTimers (node:internal/timers:500:7)
```

### DIFF
--- a/src/custom_connect.ts
+++ b/src/custom_connect.ts
@@ -10,7 +10,7 @@ export const customConnect = async (socket: net.Socket, server: http.Server): Pr
     // (whether any additional sockets are used).
 
     const asyncWrite = promisify(socket.write).bind(socket);
-    await asyncWrite.call(socket, 'HTTP/1.1 200 Connection Established\r\n\r\n');
+    await asyncWrite('HTTP/1.1 200 Connection Established\r\n\r\n');
     server.emit('connection', socket);
 
     return new Promise((resolve) => {
--- a/src/custom_connect.ts
+++ b/src/custom_connect.ts
@@ -2,7 +2,8 @@ import net from 'net';
 import type http from 'http';
 import { promisify } from 'util';
 
-const asyncWrite = promisify(net.Socket.prototype.write);
+const socket = new net.Socket();
+const asyncWrite = promisify(socket.write).bind(socket);
 
 export const customConnect = async (socket: net.Socket, server: http.Server): Promise<void> => {
     // `countTargetBytes(socket, socket)` is incorrect here since `socket` is not a target.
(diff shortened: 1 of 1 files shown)


src/custom_connect.ts
```diff
@@ -2,7 +2,8 @@ import net from 'net';
 import type http from 'http';
 import { promisify } from 'util';
 
-const asyncWrite = promisify(net.Socket.prototype.write);
+const socket = new net.Socket();
+const asyncWrite = promisify(socket.write).bind(socket);
```
jancurn, line 6: This is unnecessarily hacky, let's call `promisify` in `customConnect` function, no need to do it here (performance impact is minuscule)

jancurn (changes requested):

src/custom_connect.ts
JuroOravec: Right, I didn't realize that `customConnect` receives a Socket instance. Updated!

src/custom_connect.ts
```diff
@@ -2,15 +2,14 @@ import net from 'net';
...
-
 export const customConnect = async (socket: net.Socket, server: http.Server): Promise<void> => {
     // `countTargetBytes(socket, socket)` is incorrect here since `socket` is not a target.
     // We would have to create a new stream and pipe traffic through that,
     // however this would also increase CPU usage.
     // Also, counting bytes here is not correct since we don't know how the response is generated
     // (whether any additional sockets are used).
 
+    const asyncWrite = promisify(socket.write).bind(socket);
     await asyncWrite.call(socket, 'HTTP/1.1 200 Connection Established\r\n\r\n');
```
jancurn, line 13: Second line can be changed to simpler:
```
await asyncWrite('HTTP/1.1 200 Connection Established\r\n\r\n');
```

JuroOravec: Updated 👍

jancurn (approved):

jancurn:
@JuroOravec thanks for your help!

PR #526, opened 2023-11-02
feat: Add option to specify server's `host`

jirimoravcik:
I've added support for specifying `host` used for the `listen` call. This will enable users to e.g. listen on `localhost`, which can be more secure than the default unspecified IP address (0.0.0.0).

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,6 +39,10 @@ const server = new ProxyChain.Server({
     // Port where the server will listen. By default 8000.
     port: 8000,
 
+    // Hostname where the server will listen.
+    // optional, defaults to unspecified IP address, see e.g. https://en.wikipedia.org/wiki/IPv6_address#Unspecified_address
+    hostname: 'localhost',
+
     // Enables verbose logging
     verbose: true,
 
--- a/README.md
+++ b/README.md
@@ -39,9 +39,10 @@ const server = new ProxyChain.Server({
     // Port where the server will listen. By default 8000.
     port: 8000,
 
-    // Hostname where the server will listen.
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -39,6 +39,10 @@ const server = new ProxyChain.Server({
     // Port where the server will listen. By default 8000.
     port: 8000,
 
+    // Hostname where the server will listen.
+    // optional, defaults to unspecified IP address, see e.g. https://en.wikipedia.org/wiki/IPv6_address#Unspecified_address
```
jancurn, line 43: Let's change it to "host", as it might be also an IP address. Plus can we improve the description:

> Optional host where the proxy server will listen. If not specified, there sever listens on an unspecified IP address (0.0.0.0 in IPv4, :: in IPv6). You can use this option to limit the access to the proxy server.

barjin (approved):
Looking good, thanks :)

jancurn (commented):
Thanks Jirka. Please can we also add a (simple) unit test for the host param?

README.md
jirimoravcik: yeah, true, good point

jancurn:
Plus, can we please update the `anonymizeProxy()` function to set localhost? That will increase security by default, plus test the new option

src/anonymize_proxy.ts
```diff
@@ -56,6 +56,7 @@ export const anonymizeProxy = (
             server = new Server({
                 // verbose: true,
                 port,
+                host: 'localhost',
```
jancurn, line 59: Maybe it should be 127.0.0.1 as below? not sure ...

fnesveda (approved):

jancurn (approved):

PR #528, opened 2024-01-10
Use key/value object in place of Array for requests headers

Elyahou:
I came across an issue where the `anonymizeProxy` helper  function was not working correctly, and upstream proxy was returning 407, even if the username/password was correctly set in the proxy url.

I found out that the upstream proxy was not receiving the credential header correctly. `http` and `https` request should contain header according to [the API](https://nodejs.org/docs/latest-v14.x/api/http.html#http_http_request_options_callback).

With the proposed changes, the `anonymizeProxy` works correctly, and I no longer have errors

### DIFF
--- a/src/chain.ts
+++ b/src/chain.ts
@@ -21,7 +21,7 @@ const createHttpResponse = (statusCode: number, statusMessage: string, message =
 
 interface Options {
     method: string;
-    headers: string[];
+    headers: Record<string, string>;
     path?: string;
     localAddress?: string;
     family?: number;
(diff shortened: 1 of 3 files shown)


jancurn (approved):
Thanks for the PR, the changes look reasonable, I'm just surprised it used to work before...

jancurn (changes requested):
Just the tests are failing now...

Elyahou:
yes, so it seems that original implementation with array is to allow repeating headers, and my change break this.

Elyahou:
It seems `http` support sending an array of value to support repeating headers, so I changed the `forward` function accordingly

jancurn:
So it worked fine with the array too, instead of object? The reason why we use array of raw headers is to preserve all even duplicate headers to make the proxy transparent.

Issue #560, opened 2024-11-13
targetSocket is not closed when chain proxy replies with a non 200


arenevier:
I noticed that after closing the server with server.close(true), node did not always exit.

Actually, it fails to exit when the chain proxy has returned a non-200 response for at least one of the connection. I believe it's because targetSocket is not ended properly.

jirimoravcik:
Fix released in version `2.5.5`.

PR #561, opened 2024-11-13
close targetSocket when proxy replies with non-200 status

arenevier:
Fixes issue #560

### DIFF
--- a/src/chain.ts
+++ b/src/chain.ts
@@ -117,6 +117,8 @@ export const chain = (
                 sourceSocket.end(createCustomStatusHttpResponse(status, `UPSTREAM${statusCode}`));
             }
 
+            targetSocket.end();
+
             server.emit('tunnelConnectFailed', {
                 proxyChainId,
                 response,


jirimoravcik:
Hey @arenevier, can you please provide minimal reproduction code that wouldn't close before this fix and would close with the fix? Thank you

arenevier:
> Hey @arenevier, can you please provide minimal reproduction code that wouldn't close before this fix and would close with the fix? Thank you

Here is an example. If I send SIGINT, the server will close. But node process will hang and won't exit properly.


```
import * as ProxyChain from 'proxy-chain';     ■ Cannot find module 'proxy-chain' or its corresponding type declarations.
  
  async function main() {
  
    const server = new ProxyChain.Server({
      port: 6000,
      host: '127.0.0.1',
H     prepareRequestFunction: (params) => {     ■■ 'params' is declared but its value is never read.
            return {
              upstreamProxyUrl: "http://some.proxy.that.returns.403"
            };
      },
    });
  
    ['SIGINT', 'SIGTERM'].forEach((signal) => {
E     process.once(signal, async () => {     ■ Cannot find name 'process'. Do you need to install type definitions for node? Try `npm i --save-dev
        console.log('got signal', signal);
        // close the server gracefully
        await server.close(true);
      });
    });
  }
  
  main();
```

jirimoravcik (approved):

PR #568, opened 2025-01-15
chore: update eslint to v9, use new shared config

jirimoravcik:
This PR updates eslint to v9 along with the new flat config. We also use the new shared Apify config.


jirimoravcik:
I had to:
- Pin ubuntu in CI to 22, because with latest I was getting nasty kernel errors when launching puppeteer
- Use Node 18 instead of 16 in lint CI because of `structuredClone`, which was added in Node 17

src/utils/normalize_url_port.ts
```diff
@@ -14,7 +14,7 @@ export const normalizeUrlPort = (url: URL): number => {
         return Number(url.port);
     }
 
-    if (mapping.hasOwnProperty(url.protocol)) {
+    if (Object.prototype.hasOwnProperty.call(mapping, url.protocol)) {
```
tobice, line 17: Why so complicated? :sweat_smile: 

Would `if (url.protocol in mapping)` work?

test/tcp_tunnel.js
```diff
@@ -42,16 +43,16 @@ const closeServer = (server, connections) => new Promise((resolve, reject) => {
 
 describe('tcp_tunnel.createTunnel', () => {
     it('throws error if proxyUrl is not in correct format', () => {
-        assert.throws(() => { createTunnel('socks://user:password@whatever.com:123', 'localhost:9000'); }, /must have the "http" protocol/);
-        assert.throws(() => { createTunnel('socks5://user:password@whatever.com', 'localhost:9000'); }, /must have the "http" protocol/);
+        expectThrowsAsync(async () => { await createTunnel('socks://user:password@whatever.com:123', 'localhost:9000'); }, /must have the "http" protocol/);
```
tobice, line 46:
```suggestion
        await expect(createTunnel('socks://user:password@whatever.com:123', 'localhost:9000')).rejects.toThrow(/must have the "http" protocol/);
```

Would this work here? This is the syntax we're using e.g. in `apify-core`.

tobice (approved):
Changes look safe :+1: 

Just two suggestions how to possibly approach certain parts better.

test/tcp_tunnel.js
jirimoravcik: This is sadly not `jest` but `chai` :(

tobice: Ah, okay then :cry:

src/utils/normalize_url_port.ts
jirimoravcik: yeah, looks like it should. changed it, thanks.

fnesveda (approved):

PR #570, opened 2025-01-22
fix: set target socket max listeners to infinity

jirimoravcik:
It happens that `targetSocket` has many `.on('close', ...)` listeners and we get `MaxListenersExceededWarning`. I increase the limit to infinity so we don't see such warnings.

I'd guess this is due to the fact that some target sockets can be reused/kept alive and therefore they have more on close listeners attached from several source sockets.

### DIFF
--- a/src/chain.ts
+++ b/src/chain.ts
@@ -88,6 +88,9 @@ export const chain = (
     client.on('connect', (response, targetSocket, clientHead) => {
         countTargetBytes(sourceSocket, targetSocket);
 
+        // There may be many .on('close') listeners, so we need to increase the limit.
+        targetSocket.setMaxListeners(Infinity);
+
         if (sourceSocket.readyState !== 'open') {
             // Sanity check, should never reach.
             targetSocket.destroy();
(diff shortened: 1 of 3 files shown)


src/chain.ts
```diff
@@ -88,6 +88,9 @@ export const chain = (
     client.on('connect', (response, targetSocket, clientHead) => {
         countTargetBytes(sourceSocket, targetSocket);
 
+        // There may be many .on('close') listeners, so we need to increase the limit.
+        targetSocket.setMaxListeners(Infinity);
```
jancurn, line 92: Isn't this risky? The warning is there to help detect memory leaks, now we turn this off and if proxy starts crashing, we won't even know why. Maybe increase it too like 1000 instead?

fnesveda (changes requested):
How about attaching the handlers with with `.once('close')` instead of `.on('close')`? That might actually solve the issue.

jirimoravcik:
> How about attaching the handlers with with `.once('close')` instead of `.on('close')`? That might actually solve the issue.

Based on what would that solve the issue?

jirimoravcik:
I'll close this now as I'll be working to address this in a different PR. 

Indeed, the sockets are being reused and multiple handlers are being attached to them, which causes the warnings.

PR #571, opened 2025-01-24
fixed missing log prefix for the first request

peonone:
The connection ID log prefix for the first request is missing because `connectionId` starts from 0.


jancurn (approved):

jancurn:
Good catch, thank you

PR #577, opened 2025-03-18
Add HTTPS upstream proxy support with option to ignore proxy cert errors

patrickkfkan:
As title. E.g.:

```
const server = new Server({
  prepareRequestFunction: () => ({
    upstreamProxyUrl: 'https://user:pass@myproxy:8080',
    ignoreUpstreamProxyCertificate: true  // Useful for self-signed cert
  })
}).listen();
```
Anonymize:
```
anonymizeProxy({
  url: 'https://user:pass@myproxy:8080',
  ignoreProxyCertificate: true,
}).then((anonymizedURL) => {
  console.log(`Anonymized URL: ${anonymizedURL}`);
});
```
Create tunnel:
```
createTunnel(
  'https://user:pass@myproxy:8080',
  targetHost,
  { ignoreProxyCertificate: true }
).then((url) => {
  console.log('Tunnel endpoint:', url);
});
```

### DIFF
--- a/src/forward.ts
+++ b/src/forward.ts
@@ -25,6 +25,7 @@ interface Options {
 
 export interface HandlerOpts {
     upstreamProxyUrlParsed: URL;
+    ignoreUpstreamProxyCertificate: boolean;
     localAddress?: string;
     ipFamily?: number;
     dnsLookup?: typeof dns['lookup'];
@@ -79,10 +80,7 @@ export const forward = async (
         }
     }
 
-    const fn = origin!.startsWith('https:') ? https.request : http.request;
-
-    // We have to force cast `options` because @types/node doesn't support an array.
-    const client = fn(origin!, options as unknown as http.ClientRequestArgs, async (clientResponse) => {
+    const requestCallback = async (clientResponse: http.IncomingMessage) => {
         try {
             // This is necessary to prevent Node.js throwing an error
             let statusCode = clientResponse.statusCode!;
@@ -113,7 +111,16 @@ export const forward = async (
             // Client error, pipeline already destroys the streams, ignore.
             resolve();
         }
-    });
+    };
+
+    // We have to force cast `options` because @types/node doesn't support an array.
+    const client = origin!.startsWith('https:') ?
+        https.request(origin!, {
+            ...options as unknown as https.RequestOptions,
+            rejectUnauthorized: handlerOpts.upstreamProxyUrlParsed ? !handlerOpts.ignoreUpstreamProxyCertificate : undefined
+        }, requestCallback)
+        
+        : http.request(origin!, options as unknown as http.RequestOptions, requestCallback);
 
     client.once('socket', (socket: SocketWithPreviousStats) => {
         // Socket can be re-used by multiple requests.
--- a/src/forward.ts
+++ b/src/forward.ts
@@ -114,12 +114,12 @@ export const forward = async (
     };
 
     // We have to force cast `options` because @types/node doesn't support an array.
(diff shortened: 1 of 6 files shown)


jirimoravcik:
Hi @patrickkfkan,
thank you for your contribution!
the unit tests are sadly failing, would you be willing to take a look at that?

Thank you

patrickkfkan:
Hello! I missed the part where `options` can be `undefined` as the type definition for `createTunnel()` didn't reflect this. I've fixed this in the PR so unit tests should pass now. Could you please check again? Thanks!

jirimoravcik:
Hey, thanks for the fix. I noticed the lint is failing. Could you please rebase your PR on top of the current `master` branch? Thanks

patrickkfkan:
Done.

src/forward.ts
```diff
@@ -113,7 +111,16 @@ export const forward = async (
...
             resolve();
         }
-    });
+    };
+
+    // We have to force cast `options` because @types/node doesn't support an array.
+    const client = origin!.startsWith('https:')
+        ? https.request(origin!, {
+            ...options as unknown as https.RequestOptions,
+            rejectUnauthorized: handlerOpts.upstreamProxyUrlParsed ? !handlerOpts.ignoreUpstreamProxyCertificate : undefined,
```
jirimoravcik, line 120: I wonder why isn't `rejectUnauthorized: !handlerOpts.ignoreUpstreamProxyCertificate,` enough here? (you did that in `chain.ts`)

src/tcp_tunnel_tools.ts
```diff
@@ -19,14 +19,15 @@ const getAddress = (server: net.Server) => {
...
+    options?: {
         verbose?: boolean;
+        ignoreProxyCertificate?: boolean;
     },
     callback?: (error: Error | null, result?: string) => void,
 ): Promise<string> {
     const parsedProxyUrl = new URL(proxyUrl);
-    if (parsedProxyUrl.protocol !== 'http:') {
-        throw new Error(`The proxy URL must have the "http" protocol (was "${proxyUrl}")`);
+    if (parsedProxyUrl.protocol !== 'http:' && parsedProxyUrl.protocol !== 'https:') {
```
jirimoravcik, line 29: Could you please do something with an array `['http:', 'https:']` and not in?

jirimoravcik (changes requested):
Hey, I'm almost ready to merge this PR, just a few small nits. Could you please take a look at them? Thanks

src/forward.ts
patrickkfkan: Because in `server.ts`, the code checks for `upstreamProxyUrlParsed` before calling `chain()`. There is no such check for `forward()`.  In other words, it is possible for `forward()` to be called with `upstreamProxyUrlParsed: undefined` (notwithstanding your type definition of `ForwardOpts.upstreamProxyUrlParsed: URL`). If there is no check for this in `forward()`, and:

1. User doesn't provide an upstream proxy URL; but
2. Sets `ignoreUpstreamProxyCertificate` to `true` (user shouldn't do this in this case but we can't take this for granted)

Then `forward()` will make a direct request to `request.url` while ignoring certificate errors. This is obviously dangerous and needs to be avoided.

src/tcp_tunnel_tools.ts
patrickkfkan: Done.

jirimoravcik (approved):

jirimoravcik:
This feature has landed in `v2.5.8`. @patrickkfkan thank you again for your contribution!

PR #584, opened 2025-04-22
fix: pass headers as object in chain

yotamN:


jancurn:
This is intended behavior, please see comments in https://github.com/apify/proxy-chain/issues/585

jancurn (approved):

Issue #585, opened 2025-04-22
Using http.request undocumented behavior


yotamN:
I noticed that in the chain method within chain.ts, http.request is being used with headers as a string array instead of an object. This behavior is undocumented and could potentially cause issues in the future. Currently, OpenTelemetry auto-instrumentation is affected because it expects headers to be an object.

Is there a specific reason for using an array for headers? If not, I have opened [a pull request](https://github.com/apify/proxy-chain/pull/584) to address this issue.

jancurn:
The reason is that some servers or clients duplicate HTTP headers and we want to proxy the headers as they are, to reduce a chance of blocking. For details, see https://nodejs.org/docs/latest/api/http.html#responsewriteheadstatuscode-statusmessage-headers, where this behavior is documented

yotamN:
I understand your point. However, in the case of a CONNECT request, isn't the order of headers only relevant to the proxy server itself? This means that I cannot change it in forward mode. Did your research conclude that this is the only way to order the headers?

Additionally, regarding the documentation, you referred to the documentation for `response.writeHead`. The documentation for `http.request` still does not mention the handling of headers as an array.

I will proceed to fix this in OpenTelemetry as well, since you are likely not the only library affected. However, it is important to note that this behavior could technically change in any future version.

jancurn:
Oh you're right, I thought this is an external interface, but here in this case it make sense to change this as this is internal usage. Thanks for your help.

yotamN:
Thank you.
I also [opened an issue](https://github.com/nodejs/node/issues/57986) with NodeJS and it seems someone already opened a PR in there as well.

jirimoravcik:
Fixed in https://github.com/apify/proxy-chain/pull/584

tonybruess:
It seems like `forward` is also passing headers as an array which is breaks Charlies and Brightdata proxies. Can I open a PR to fix?

jancurn:
> breaks Charles and Brightdata proxies

How does it break them? You mean remote proxies or some local ones?

tonybruess:
Example of array headers breaking charles:

<img width="932" alt="Image" src="https://github.com/user-attachments/assets/f29e1fa0-cca0-4d6e-9250-4eb6af1a3a7b" />

Versus passing headers as an object:

<img width="1025" alt="Image" src="https://github.com/user-attachments/assets/7e8d917b-a720-4420-9fbb-930defb0af72" />

yotamNimble:
[...]

PR #588, opened 2025-05-20
fix: pass headers as object in forward and forwardSocks

tonybruess:
Following up #585 and #584, headers should be passed as an object. This is the documented and expected type. Passing as an array breaks Charles and doesn't work with brightdata proxies.

### DIFF
--- a/src/utils/valid_headers_only.ts
+++ b/src/utils/valid_headers_only.ts
@@ -1,3 +1,4 @@
+import type { IncomingHttpHeaders, OutgoingHttpHeaders } from 'node:http';
 import { validateHeaderName, validateHeaderValue } from 'node:http';
 
 import { isHopByHopHeader } from './is_hop_by_hop_header';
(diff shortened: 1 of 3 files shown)


src/utils/valid_headers_only.ts
```diff
@@ -1,22 +1,27 @@
...
+export const validHeadersOnly = (rawHeaders: IncomingHttpHeaders): OutgoingHttpHeaders => {
+    const result: OutgoingHttpHeaders = {};
 
     let containsHost = false;
 
-    for (let i = 0; i < rawHeaders.length; i += 2) {
-        const name = rawHeaders[i];
-        const value = rawHeaders[i + 1];
-
+    for (const [name, value] of Object.entries(rawHeaders)) {
```
tonybruess, line 14: Unclear to me if we still need this. Are `request.headers` already validated?

tonybruess:
Also, both both of these data types support arrays of strings for duplicate headers?

tonybruess:
Brightdata proxies aren't working and I'm happy to provide credentials if there's a private space I can share them

PR #597, opened 2025-09-24
Emit event `tunnelError` when the tunnel has an error

djxy:
When an error happens in the chaining it is only logged, but not emitted. I created the event `tunnelError` that emit the error related to the tunnel.


bliuchak:
@djxy I'll take a look on them tomorrow

bliuchak:
Looks like there are some issues with unit tests. I'm gonna fix them at https://github.com/apify/proxy-chain/issues/599

Issue #599, opened 2025-09-26
Unit tests are failing


bliuchak:
Investiage and fix why unit tests are either stuck forever or fail.

PR #600, opened 2025-09-26
fix: can't run unit tests in docker

bliuchak:
This PR fix Dockerfile and allows unit tests to run inside a container gracefully. ~~Also it specifies a node.js version `18` into `.nvmrc` file to ensure compatibility.~~ Discussed later with Jirka that it's better not to have this file. All the tests must be fixed and should be compatible with newer versions of node all the time.

Resolves: #599


jirimoravcik:
SOCKS test is failing in CI

bliuchak:
@jirimoravcik should be fixed now.

jirimoravcik:
> I've already checked node 22 and tests are failing there due to few breakable changes (e.g. node 22 don't allow us to response with status code `55`). I'd open a new issue for it if you don't mind.

Yes, that's fine

.nvmrc
jirimoravcik, line 1: I would remove this file. As this is a library, I don't see any benefit in fixing the version. E.g. see https://github.com/apify/crawlee, there's no `.nvmrc`

jirimoravcik (changes requested):

.nvmrc
bliuchak: Removed.

PR #602, opened 2025-10-08
feat: implement https server support

bliuchak:
This PR implements HTTPS server support to proxy-chain.

Also fixed:
- a datarace for `error` event when we might log same events
- fix for usage statistics trackings

Otherwise my changes should be fully compatible with HTTP server and all the handlers.


src/server.ts
```diff
@@ -243,11 +358,32 @@ export class Server extends EventEmitter {
...
+
+            // Log errors only if there are no user-provided error handlers
+            if (this.listenerCount('error') === 0) {
                 this.log(socket.proxyChainId, `Source socket emitted error: ${err.stack || err}`);
             }
         });
+
+        // Check once per connection for socket._parent availability.
+        if (this.serverType === 'https') {
+            const rawSocket = socket._parent;
```
jirimoravcik: It may be worth asking in https://github.com/nodejs/node about the safety of using this private property.

src/server.ts
```diff
@@ -169,11 +227,44 @@ export class Server extends EventEmitter {
...
+                honorCipherOrder: true, // Server chooses cipher (prevents downgrade attacks)
+                ...options.httpsOptions, // User options override defaults
+            };
+
+            this.server = https.createServer(secureDefaults);
+            this.serverType = 'https';
+        } else {
+            this.server = http.createServer();
+            this.serverType = 'http';
+        }
```
jirimoravcik: Maybe validate if `options.serverType` is one of `http`, `https`? It would make it consistent with the type. I'd also set it to `http` by default in the constructor parameter. That way you could just do `this.serverType = options.serverType`

src/server.ts
```diff
@@ -613,16 +749,33 @@ export class Server extends EventEmitter {
...
-            srcTxBytes: socket.bytesWritten,
-            srcRxBytes: socket.bytesRead,
-            trgTxBytes: targetStats.bytesWritten,
-            trgRxBytes: targetStats.bytesRead,
+        return {
+            srcTxBytes, // HTTP: app only, HTTPS: total (app + TLS overhead)
+            srcRxBytes, // HTTP: app only, HTTPS: total (app + TLS overhead)
+            trgTxBytes: targetStats?.bytesWritten,
+            trgRxBytes: targetStats?.bytesRead,
         };
```
jirimoravcik: Just an idea. Why don't we store the `_parent` socket in `this.connections`? Looking at the logic here in `getConnectionsStats`, you just fall back to the original socket for stats. That makes me question why would we ever want to use the TLS socket for connection tracking?
That gives me another idea, if you do `this.server.on('connection')` for HTTPS you should be able to reach the original Socket without using `_parent`, right?

jirimoravcik (changes requested):
Found a few more points for discussion

src/server.ts
bliuchak: Good point. Gonna add this.

src/server.ts
jirimoravcik: Ok, I agree

src/server.ts
```diff
@@ -243,11 +358,32 @@ export class Server extends EventEmitter {
...
         // See https://github.com/apify/proxy-chain/issues/53
         socket.on('error', (err) => {
-            // Handle errors only if there's no other handler
-            if (this.listenerCount('error') === 1) {
+            // Prevent duplicate error handling for the same socket
+            if (socket.proxyChainErrorHandled) return;
+            socket.proxyChainErrorHandled = true;
+
+            // Log errors only if there are no user-provided error handlers
+            if (this.listenerCount('error') === 0) {
```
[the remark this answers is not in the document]
jirimoravcik: So it was broken before, right?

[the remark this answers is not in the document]
bliuchak: Yes, it was broken before. But I was wrong in my assumptions.

Original code should check `socket.listenerCount('error') === 1` instead for `this` (server). Because socker and server error events aren't interconnected.

Also, we're attaching additional error listeners in our handlers (e.g. [direct](https://github.com/apify/proxy-chain/blob/master/src/direct.ts#L106)).

So, basically the flow should be like that:
1. `onConnection()` - attach out early handler - `socket.listenerCount('error') === 1`
2. `onReqeust() / onConnect()` - prepareRequestHandling
3. `direct()` function called - attaches sourceSocket error handler
4. Now `socket.listenerCount('error') === 2`

If error occurs after step 3: then the direct() handler will log instead our early handler.

If error occurs between steps 1-2: then our early handler will log it.

I'm gonna prepare fix for this.

scutuatua-crypto (approved):

Issue #606, opened 2025-11-01
Missing TLS overhead in byte statistics when using HTTPS upstream proxy


bliuchak:
When using an HTTPS upstream proxy, connection statistics (trgRxBytes/trgTxBytes) only count application-layer bytes and miss TLS overhead.

This affects both handlers:
- forward.ts (HTTP requests via HTTPS upstream)
- chain.ts (CONNECT tunneling via HTTPS upstream)

Expected behavior: Byte counts should include all TCP bytes by accessing the `_parent` socket of TLS connection.

```bash
npm run test test/https_upstream_tls_overhead.js
...
  2 passing (1s)
  2 failing

  1) forward.ts TLS overhead with HTTPS upstream
       should demonstrate missing TLS overhead with HTTPS upstream proxy:

      counted trgRxBytes 171 not equal to actual upstream count 2300
      + expected - actual

      -171
      +2300

      at Context.<anonymous> (test/https_upstream_tls_overhead.js:230:44)

  2) chain.ts TLS overhead with HTTPS upstream
       should demonstrate missing TLS overhead with HTTPS upstream proxy (chain handler):

      counted trgRxBytes 2339 not equal to actual upstream count 4512
      + expected - actual

      -2339
      +4512

      at Context.<anonymous> (test/https_upstream_tls_overhead.js:495:44)

```

bliuchak:
Closing this in favor of https://github.com/apify/proxy-chain/issues/631

PR #607, opened 2025-11-01
test: missing tls overhead for https upstream

bliuchak:
These are few tests that proofs missing tls overhead when https upstream is used. There should be 2 test cases which will be failing.

Checklist:
- [ ] Create a fix
- [x] Write tests that reproduces the problem
- [ ] Make sure that failing tests are passing after fix :)

See #606


bliuchak:
Closing this PR due to low priority of this task.

PR #618, opened 2025-11-10
perf: do not publish `tsbuildinfo` file to npm

barjin:
Shrinks the package size by 13kB compressed or 35kB uncompressed space.

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,7 +28,8 @@
     },
     "homepage": "https://blog.apify.com/how-to-make-headless-chrome-and-puppeteer-use-a-proxy-server-with-authentication-249a21a79212",
     "files": [
-        "dist"
+        "dist",
+        "!**/*.tsbuildinfo"
     ],
     "scripts": {
         "build:watch": "tsc -w",


barjin:
It's generated in `dist/`, so it's already `.gitignore`d. Apparently, the `files` field has precedence over `.npmignore`, so adding it there didn't help either.

I suppose we could edit `tsconfig` to generate it in the repo root instead of `dist` (the current tsconfig is missing `rootDir`), but I took the path of least resistance (not sure about the other consequences of such change). wdyt?

B4nan (approved):
I just wonder why this is not needed in crawlee or other repos, where we only gitignore the file.

barjin:
> why this is not needed in crawlee

It seems it is :)

image

By default, this file goes to `outDir`. If `rootDir` is set (see e.g. [@crawlee/utils)](https://github.com/apify/crawlee/blob/master/packages/utils/tsconfig.build.json), then it goes in `rootDir`'s and `tsconfig`'s common parent (or something like that). More info in [docs](https://www.typescriptlang.org/tsconfig/#tsBuildInfoFile).

B4nan:
oh, i only checked the core package where its not present.

PR #619, opened 2025-11-12
Add HTTP agent support for connection pooling

paveldudka:
## Problem

Currently, proxy-chain creates a new TCP connection to the upstream proxy for every request. This causes:
- **New IP addresses per request** when using residential/rotating proxies (breaks session-based workflows)
- **Performance overhead** from repeated TCP handshakes
- **No connection reuse** despite HTTP/1.1 keepalive support

## Solution

Add optional `httpAgent` and `httpsAgent` parameters to `prepareRequestFunction` result, enabling users to provide persistent HTTP agents with `keepAlive: true`.

## Changes

- Add `httpAgent?` and `httpsAgent?` to `PrepareRequestFunctionResult` type
- Thread agents through handler pipeline (`server.ts`, `forward.ts`, `chain.ts`)
- Pass agents to `http.request()` and `https.request()` calls

## Benefits

- ✅ **Sticky IP** - Same connection = same IP from residential proxy pool
- ✅ **Better performance** - Connection reuse eliminates handshake overhead
- ✅ **Backward compatible** - Agents are optional, existing code works unchanged
- ✅ **Preserves all existing features** - `getConnectionStats()`, events, etc. still work

## Usage

```javascript
const http = require('http');
const { Server } = require('proxy-chain');

const httpAgent = new http.Agent({
  keepAlive: true,
  keepAliveMsecs: 30000
});

const server = new Server({
  prepareRequestFunction: () => ({
    upstreamProxyUrl: 'http://proxy.example.com:8080',
    httpAgent,  // Reuse connections
  }),
});
```

## Testing

Added comprehensive test suite in `test/http-agent.js`:
- Agent acceptance and threading
- Connection pooling verification
- Backward compatibility
- `getConnectionStats()` preservation
- HTTP vs HTTPS agent selection

All tests passing ✅

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
     "name": "proxy-chain",
-    "version": "2.5.9",
+    "version": "2.6.0",
     "description": "Node.js implementation of a proxy server (think Squid) with support for SSL, authentication, upstream proxy chaining, and protocol tunneling.",
     "main": "dist/index.js",
     "keywords": [


test/http-agent.js
```diff
@@ -0,0 +1,268 @@
...
+                    upstreamProxyUrl: `http://localhost:${upstreamProxyPort}`,
+                    httpAgent,
+                };
+            },
+        });
+
+        await mainProxyServer.listen();
+
+        // Make multiple requests
+        for (let i = 0; i < 3; i++) {
```
bliuchak: Nice to have: would be good to actually use concurrent reqeusts here to make test more challenging

test/http-agent.js
```diff
@@ -0,0 +1,268 @@
...
+        });
+
+        expect(agentsPassed).to.be.true;
+
+        // Cleanup agents
+        httpAgent.destroy();
+        httpsAgent.destroy();
+    });
+
+    it('reuses connections with keepAlive agents (sticky IP simulation)', async () => {
```
bliuchak: What about instead of verifing implementation (hooks into 'free' event, inspect socket implementation details like remoteAddress and port, compare socketIds) to veirfy an observable effect of connection pooling (behavior)?

What we can do instead? Counts actual connections at the upstream proxy server. With this we can verify that fewer connections are made than requests (observable behavior of pooling).

What do you think?

bliuchak (changes requested):
Hi @paveldudka! Big thanks for your contribution and quick fixes 🎉 

Here are few issues from my side that should be addressed before merge:

1. Missing test coverage for CONNECT tunneling (chain handler). Currently all targets in tests are HTTP. Please add a tests for HTTPS target.
2. Missing HTTPS upstream proxy tests (your last test claims that it verifies HTTPS agent alognside with HTTP, however no code there verifies that httpsAgent is actually used). It would be good to have this covered as separate test.
3. Document SOCKS limitation. Agents don't work with SOCKS upstream proxies (forward_socks.ts, chain_socks.ts). Would be good to add some JSDoc or mention it in README.
4. Nice to have: Add usage example of agents into README.

Also, you can treat my comments for code itself as nice to have. Overall implementation looks good. If you'll manage them - good, if not - also good ;)

Let me know if you'll need any clarification.

test/http-agent.js
paveldudka: that's a great idea. Lemme dwell on it

paveldudka:
@bliuchak , reworked tests to move away from testing implementation details
Testing httpsServer is quite tricky as I couldn't find a good way to simulate a fully functional https test server. Kinda worked around it by testing that connection is being attempted. LMK what you think

bliuchak:
@paveldudka , thanks for adding examples and document limitation 👏 

However the main issue is still there - lack of test coverage for HTTPS upstream proxies (both forward and chain mode).

- HTTPS upstream proxy in chain mode.
Current tests cover HTTPS targets through HTTP upstream but missing HTTPS targets through HTTPS upstream.
- HTTPS upstream proxy in forward mode.
As you wrote there - current tests uses non-existent proxy URL which is only a smoke test (to verify that code doesn't crash). However what we also need to have is a functional test with real HTTPS upstream proxy and connection pooling verification.

Could you please add these two tests?

paveldudka:
> @paveldudka , thanks for adding examples and document limitation 👏
> 
> However the main issue is still there - lack of test coverage for HTTPS upstream proxies (both forward and chain mode).
> 
> * HTTPS upstream proxy in chain mode.
>   Current tests cover HTTPS targets through HTTP upstream but missing HTTPS targets through HTTPS upstream.
> * HTTPS upstream proxy in forward mode.
>   As you wrote there - current tests uses non-existent proxy URL which is only a smoke test (to verify that code doesn't crash). However what we also need to have is a functional test with real HTTPS upstream proxy and connection pooling verification.
> 
> Could you please add these two tests?

@bliuchak , If I understand the pipeline correctly, in forward mode (HTTP requests), the proxy terminates each request, allowing the agent to pool TCP connections for subsequent requests. In chain mode (HTTPS via CONNECT), each CONNECT establishes a long-lived tunnel for the client's TLS connection to the target - these tunnels can't be "reused" for different targets, so each requires its own upstream proxy connection. 
The httpsAgent still provides value by managing these connections (maxSockets, timeout handling, etc.). 
So technically with https upstream we can't really have connection pooling in a traditional sense.

But at this point Im hitting my domain knowledge limitations in this space, so will defer to you to decide how to proceed. Locally I have a forked version of proxy-chain and we use http upstream, so in the short term, we are unblocked

bliuchak (approved):

PR #638, opened 2026-01-12
feat: update minimal supported version of Node.js to v20

### LINKED ISSUE #601: Upgrade to newer version of node.js
Currently project works under Node 18 which reached its EOL on April 30, 2025. It will no longer receive security updates or maintenance releases.

Plan it to upgrade to version 22 or 24 (current LTS).

Note: this won't be an easy task since there are brekable changes.

Refs: https://nodejs.org/en/blog/announcements/node-18-eol-support

bliuchak:
This PR upgrades minimal supported version of Node.js to v20.

## Why Node.js 20?

I think for us it makes sense to support last 3 LTS versions of node. As of today they are: 24, 22 and 20.

Also I'll keep tests at Github Actions for Node.js 16, 18 and 20 (will be added) to make sure we're not breaking too much things. At least right now. However this might change in next major release.

## Migration Flow

Migration from Node.js 18 to 20 wasn't complicated. Here are the main issues which we've faced:
- Node.js 20 enables `keep-alive` by default.  This should be turned off in tests explicitly.
- Some tests should be skipped since Node.js became more strict :+1: 
- For anonymized proxy script Node.js 20 throw `socket hang up` instead of `ECONNREFUSED` for some special cases

### Overview

There was 2242 tests cases for Node.js v18. Currently there are 2191 (minus 51 test case) for Node.js v20. See details below:

1. "ignores non-standard server HTTP headers" - Skipped when `!useUpstreamProxy`:                                                                                                                                    
    - Main proxy configurations: 2 (http/https) × 2 (ssl) × 4 (auth) = 16 configs                                                                                                                                    
    - Custom response suites: 2 x 2 x 4 = 16 configs                                                                                                                                                                 
    - Direct tests (HTTP/HTTPS -> Target): 2 configs                                                                                                                                                                 
    - Total: 34 tests
2. "gracefully fails on invalid HTTP status code" - Skipped when `!useUpstreamProxy && !useSsl`:                                                                                                                     
    - Main proxy configurations: 2 x 1 x 4 = 8 configs                                                                                                                                                               
    - Custom response suites: 2 x 1 x 4 = 8 configs                                                                                                                                                                  
    - Direct tests: 1 config (HTTP -> Target only)                                                                                                                                                                   
    - Total: 17 tests
  
Grand total: 34 + 17 = 51 skipped tests

### Why these tests are skipped

Both tests verify the proxy's behavior when receiving malformed HTTP responses from upstream servers:
1. "ignores non-standard server HTTP headers" - tests handling of headers with spaces in names (e.g., Invalid Header With Space: value) which violate HTTP RFC
3. "gracefully fails on invalid HTTP status code" - tests handling of invalid status codes (like HTTP/1.1 55 OK)

In Node.js 20, the `--insecure-http-parser` flag no longer relaxes HTTP client parsing for these edge cases. The HTTP client throws parse errors before the proxy can even process them. This is a Node.js security hardening change, not a proxy bug.

Close #601

### DIFF
--- a/.github/workflows/check.yaml
+++ b/.github/workflows/check.yaml
@@ -14,7 +14,7 @@ jobs:
 
     strategy:
       matrix:
-        node-version: [14, 16, 18]
+        node-version: [16, 18, 20]
 
     steps:
       - name: Checkout repository
@@ -30,7 +30,7 @@ jobs:
 
       # Lint only on the latest Node.js version to save time.
       - name: Lint code
-        if: ${{ matrix.node-version == 18 }}
+        if: ${{ matrix.node-version == 20 }}
         run: npm run lint
 
       - name: Add localhost-test to Linux hosts file
--- a/.github/workflows/check.yaml
+++ b/.github/workflows/check.yaml
@@ -14,7 +14,7 @@ jobs:
 
         strategy:
             matrix:
-                node-version: [16, 18, 20]
+                node-version: [20, 22, 24]
 
         steps:
             - name: Checkout repository
--- a/.github/workflows/check.yaml
+++ b/.github/workflows/check.yaml
@@ -3,38 +3,38 @@
 name: Check
 
 on:
-  pull_request:
-  workflow_call:
+    pull_request:
+    workflow_call:
 
 jobs:
-  lint_and_test:
-    name: Lint & Test
-    if: ${{ !contains(github.event.head_commit.message, '[skip ci]') }}
-    runs-on: ubuntu-22.04
+    lint_and_test:
+        name: Lint & Test
+        if: ${{ !contains(github.event.head_commit.message, '[skip ci]') }}
+        runs-on: ubuntu-22.04
 
-    strategy:
-      matrix:
-        node-version: [16, 18, 20]
+        strategy:
+            matrix:
+                node-version: [16, 18, 20]
 
-    steps:
-      - name: Checkout repository
-        uses: actions/checkout@v6
+        steps:
+            - name: Checkout repository
+              uses: actions/checkout@v6
 
-      - name: Use Node.js ${{ matrix.node-version }}
-        uses: actions/setup-node@v6
-        with:
-          node-version: ${{ matrix.node-version }}
+            - name: Use Node.js ${{ matrix.node-version }}
+              uses: actions/setup-node@v6
+              with:
+                  node-version: ${{ matrix.node-version }}
 
-      - name: Install Dependencies
-        run: npm install
+            - name: Install Dependencies
+              run: npm install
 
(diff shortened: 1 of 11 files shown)


.github/workflows/check.yaml
```diff
@@ -14,7 +14,7 @@ jobs:
 
     strategy:
       matrix:
-        node-version: [14, 16, 18]
+        node-version: [16, 18, 20]
```
jirimoravcik: Why are we running checks on unsupported Node.js versions? Shouldn't this be `[20, 22, 24]`?

jirimoravcik (approved):

package.json
```diff
@@ -44,7 +44,7 @@
         "lint:fix": "eslint . --fix"
     },
     "engines": {
-        "node": ">=14"
+        "node": ">=20"
```
danpoletaev, line 47: Shouldn't minimal node version be 16?
It should just warn user to use node 20, right?

danpoletaev: Let's also update `@types/node`

test/Dockerfile
```diff
@@ -1,4 +1,4 @@
-ARG NODE_IMAGE=node:18.20.8-bookworm
+ARG NODE_IMAGE=node:20.19.6-bookworm
```
danpoletaev, line 1: We need to also update `scripts/test-docker-all.sh`, since there're hardcoded node versions

danpoletaev (changes requested):
Looks good!

Left few suggestions that needs to be reviewed before merging

.github/workflows/check.yaml
bliuchak: Good point. For some reason I kept this because all tests passed. But, you're probably right to drop old versions.

package.json
bliuchak: The idea was to set minimal engine version to current oldest supported LTS version of Node. As of today it's Node 20.

Node 18 both active and security support period has ended.

test/Dockerfile
bliuchak: You're right. Gonna fix this.

.github/workflows/check.yaml
bliuchak: Fixed.

test/Dockerfile
bliuchak: Fixed

package.json
bliuchak: > Let's also update @types/node

Good point. Upgraded.

danpoletaev (approved):
Nice ❤️

lewis-wow (approved):

jancurn:
Hey guys, any blockers here? Let's ship this

PR #659, opened 2026-05-19
chore: migrate from npm to pnpm 10

### LINKED ISSUE #658: Migrate to pnpm
We need to enable features like minimum release age enforcement to prevent us from being hit by supply chain attacks. NPM does not support this, therefore we need to switch to PNPM.

B4nan:
Closes #658.

The driver is `pnpm-workspace.yaml`'s `minimumReleaseAge: 1440` — a 24-hour quarantine on newly published versions blunting supply-chain attacks. npm has no equivalent; this is why the issue requested the migration.

## Two commits, intentionally separated
1. **Migrate to pnpm 11** (`b789ce6`).
2. **Downgrade to pnpm 10** (`3396fdc`) — keeps Node 20 in the supported matrix. pnpm 11 requires Node ≥22, which would force-drop Node 20. Kept separate so it can be reverted cleanly the moment we drop Node 20 support.

## Lockfile / package manager
- `packageManager: pnpm@10.33.4` + matching `devEngines.packageManager` with `onFail: error`.
- `engines.node` stays at `>=20.11` (matching the previous floor).
- **Commit `pnpm-lock.yaml`.** The previous setup had `package-lock.json` in `.gitignore`, so CI was effectively unpinned — every `npm install` re-resolved deps from the registry. That's exactly the surface this PR closes.
- `prepublishOnly`: `npm run build` → `pnpm run build`.
- `before-beta-release.cjs`: `npm show` → `pnpm view`. npm CLI is blocked under devEngines `onFail: error`; `pnpm view` is functionally equivalent.

> **Caveat (pnpm 10 + devEngines):** pnpm 10 shells out to npm for `pnpm config` / `pnpm version` / `pnpm pkg`, so those subcommands break under `onFail: error`. We don't invoke any of them in scripts or CI (`pnpm publish`, `pnpm view`, `pnpm install`, `pnpm run`, `pnpm deploy` are all native and safe), so the guard stays useful. This caveat goes away on pnpm 11.

## `pnpm-workspace.yaml`
- `minimumReleaseAge: 1440` (24 h) with `@apify/*` excluded.
- `allowBuilds`: `puppeteer` (Chromium download, required by e2e), `esbuild` (platform binary, required by tsx); `core-js` denied (postinstall is just a donation banner).
- `strictDepBuilds: false` so a new build-script transitive doesn't fail `--frozen-lockfile` installs.

## CI (`check.yaml`, `release.yaml`)
- Every install step collapses to:
  ```yaml
  - uses: apify/actions/pnpm-install@v1.1.2
  ```
- Test matrix stays `[20, 22, 24]`. Node 26 is intentionally **NOT** added: a pre-existing tsx + yargs (16.x CJS bin) + Node 26 ESM-loader incompatibility breaks `mocha` under the repo's `node-option: ["import=tsx"]` config (reproducible on `master` with plain `npm install`, so not a migration regression — fix belongs in a separate change).
- `release.yaml`: `npm publish` → `pnpm publish --tag <…> --no-git-checks`. npm publish would be rejected under devEngines `onFail: error`; pnpm publish honours the dist-tag and inherits OIDC auth from `setup-node`'s `registry-url`.

## Notes for review
- The `github-registry-token` input on `apify/actions/pnpm-install` is **not** passed — this repo has no `@apify-packages/*` deps (only public-registry packages).
- The previous npm-based setup never committed a lockfile, so `--frozen-lockfile`-style strictness is new here. Renovate / dependency updates will start writing to `pnpm-lock.yaml` going forward.
- Verified locally: `pnpm install` clean on both Node 20 and Node 24; `pnpm run build`, `pnpm run lint`, `pnpm run test:unit` (4 passing) all green.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


package.json
```diff
@@ -56,8 +56,16 @@
         "lint:fix": "eslint . --fix"
     },
     "engines": {
-        "node": ">=20.11"
+        "node": ">=22"
```
B4nan: I see we kept compact with node 20 in the just-released major version, so I guess we'll need to go back to pnpm v10 (v11 requires node 22)...

(or we just remove the CI checks and keep pnpm v11)

bliuchak (approved):

PR #666, opened 2026-05-29
test(e2e): upgrade puppeteer to v25 to fix broken e2e navigation

B4nan:
## Problem

All 52 puppeteer-backed e2e cases started failing wholesale — every one timing out on navigation, ballooning the `E2E tests (Node.js 24)` job from ~2 min to ~57 min ([run 26636921133](https://github.com/apify/proxy-chain/actions/runs/26636921133/job/78499498016)).

Root cause is **dependency rot triggered by a runner-image bump**, not flakiness:

- Last green master run used GitHub image `ubuntu24/20260518.149`; the failing runs use `ubuntu24/20260525.161`.
- puppeteer was pinned at `19.11.1`, whose bundled Chromium (~115, mid-2023) **launches but can no longer navigate** on the newer OS. `curl` through the same proxy still passes, confirming the proxy itself is fine and the problem is the stale browser.

## Fix

Upgrade puppeteer `19.11` → `25.x`, which ships a current Chromium built for the new image. Two renamed launch options:

- `ignoreHTTPSErrors` → `acceptInsecureCerts`
- `headless: 'new'` → `headless: true` (the `'new'` value was removed; modern headless is the default `true`)

This also drops an earlier launch/navigate-retry + per-step-timeout experiment that was based on a wrong root cause (assumed flaky spawns) — it didn't help and tripled the failing run's wall-clock time by retrying a browser whose navigation could never succeed.

Verified locally with puppeteer 25 that the new launch options launch and navigate, both directly and through an HTTP proxy.

### DIFF
--- a/package.json
+++ b/package.json
@@ -85,7 +85,7 @@
         "nyc": "^15.1.0",
         "portastic": "^1.0.1",
         "proxy": "^1.0.2",
-        "puppeteer": "^19.6.3",
+        "puppeteer": "^25.1.0",
         "request": "^2.88.2",
         "rimraf": "^4.1.2",
         "sinon": "^13.0.2",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -73,8 +73,8 @@ importers:
         specifier: ^1.0.2
         version: 1.0.2
       puppeteer:
-        specifier: ^19.6.3
-        version: 19.11.1(typescript@5.9.3)
+        specifier: ^25.1.0
+        version: 25.1.0
       request:
         specifier: ^2.88.2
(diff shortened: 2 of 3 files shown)


B4nan:
@bliuchak I'd reconsider running a CI check that takes ~30 minutes, we usually run e2e tests on master on a schedule once a day, it makes little to no sense to block PRs with this.

B4nan:
Huh, so this was all about a dated puppeteer, this PR bumps it to resolve the flakiness and it now runs in ~3 minutes, so let's leave it in PR checks, that's completely reasonable.

bliuchak (approved):

PR #667, opened 2026-05-31
Fix invalid certificate error in tcp_tunnel_test by using wait helper

MatrixNeoKozak:
The tcp_tunnel test file was missing the wait helper needed by closeServer, causing a ReferenceError. Added the missing arrow function (identical to the one in server.js) to make the test runnable again.


jancurn:
How does this work? `wait` is never referenced

Issue #660, opened 2026-05-21
createTunnel binds the relay on all interfaces instead of localhost


MN755:
## Summary
`src/tcp_tunnel_tools.ts` exposes `createTunnel()` as a localhost helper, but it actually calls `server.listen(0)` with no host and returns the resulting unspecified bind address.

## Repro
1. Call `createTunnel('http://user:pass@proxy.example.com:8000', 'service.example.com:356')`.
2. Check the implementation in `src/tcp_tunnel_tools.ts`:
   - the documented `hostname` / `port` options are not present in the function signature
   - `server.listen(0)` is called without a host
   - `getAddress()` returns the bound address verbatim
3. On current Node, an omitted host binds the listener to the unspecified address (`::` / `0.0.0.0`), not to loopback.

## Expected
`createTunnel()` should default to a loopback-only listener, consistent with the README's `hostname: localhost` contract and the `localhost:56836` example.

## Actual
The helper creates a listener on an unspecified address and returns that externally reachable endpoint.

## Likely Cause
`src/tcp_tunnel_tools.ts`, `createTunnel()` and `getAddress()`

## Source / Trigger
The caller controls `proxyUrl` and `targetHost`, then `createTunnel()` opens a local relay that forwards every incoming TCP connection through the configured upstream proxy to that target.

## Control Gap
The code never supplies a loopback host to `server.listen(...)`, and the documented `hostname` / `port` options from `README.md` are no longer implemented in the exported API.

## Impact
This turns a helper that looks localhost-scoped into an all-interfaces tunnel. If the caller points it at an internal target or at an authenticated upstream proxy, any peer that can reach the caller's machine can reuse that relay and reach the tunneled destination unexpectedly.

## Fix Direction
Bind `createTunnel()` to loopback by default (`127.0.0.1` / `::1` or `localhost`), restore or implement the documented `hostname` / `port` options, and add a regression test that fails if the returned tunnel endpoint is not loopback-scoped by default.

VojtaM39:
Issue closed by: https://github.com/apify/proxy-chain/pull/675

Thank you for reporting this!

PR #675, opened 2026-08-24
fix: don't listen on every interface in createTunnel

VojtaM39:
`createTunnel()` called `server.listen(0)` without a host, so the tunnel bound the
  unspecified address (`::` / `0.0.0.0`). Any peer that could reach the host could tunnel
  to `targetHost` using those credentials.

  GHSA-5vwf-g8jp-pgj3, #660.

  ## Changes:
  - `createTunnel()` listens on `127.0.0.1` by default. The `port` and `hostname` options
    documented in the README work again — they were dropped from the implementation in v2.0.0.
    Pass `hostname: '0.0.0.0'` to opt back into the old binding.
  - Warns when bound to a non-loopback address.
  - Tests for the listener address.
  - additional change - anonymizeProxy has optional port and defaults to 0 (random port)

  Behavior change: the returned endpoint is `127.0.0.1:<port>` instead of `0.0.0.0:<port>`.
  
  
  I have also pushed a branch `tmp/tunnel-connect-fix-with-repro` with repro, that demonstrates that the fix works.

### DIFF
--- a/README.md
+++ b/README.md
@@ -566,7 +566,7 @@ The optional `options` parameter is an object with the following properties:
 
 The result of the function is a local endpoint in a form of `hostname:port`,
 where `hostname` is the address the tunnel actually bound to. IPv6 addresses are
-bracketed, e.g. `[::1]:56836`. Pass this exact string to `closeTunnel()`.
+bracketed, e.g. `[::1]:56836`.
 
 All TCP connections made to the local endpoint will be tunneled through the proxy to the target host and port.
 For example, this is useful if you want to access a certain service from a specific IP address.
--- a/README.md
+++ b/README.md
@@ -560,18 +560,23 @@ specified by the `proxyUrl` parameter.
 The optional `options` parameter is an object with the following properties:
 - `port: Number` - Enables specifying the local port to listen at. By default `0`,
    which means a random port will be selected.
-- `hostname: String` - Local hostname to listen at. By default `localhost`.
+- `hostname: String` - Local hostname to listen at. By default `127.0.0.1`.
 - `ignoreProxyCertificate` - For HTTPS proxy, ignore certificate errors in proxy requests. Useful for proxy with self-signed certificate. By default `false`.
 - `verbose: Boolean` - If `true`, the functions logs a lot. By default `false`.
 
-The result of the function is a local endpoint in a form of `hostname:port`.
+The result of the function is a local endpoint in a form of `hostname:port`,
+where `hostname` is the address the tunnel actually bound to. IPv6 addresses are
+bracketed, e.g. `[::1]:56836`. Pass this exact string to `closeTunnel()`.
+
 All TCP connections made to the local endpoint will be tunneled through the proxy to the target host and port.
 For example, this is useful if you want to access a certain service from a specific IP address.
 
-The tunnel should be eventually closed by calling the `closeTunnel()` function.
+The tunnel does not authenticate its clients and forwards the `proxyUrl`
+credentials upstream, which is why it listens on `127.0.0.1` by default. Only
+pass a non-loopback `hostname` if you restrict access to it by other means -
+doing so emits a `ProxyChainSecurityWarning` via `process.emitWarning()`.
 
-The `createTunnel()` function returns a promise that resolves to a String with
-the path to the local endpoint.
+The tunnel should be eventually closed by calling the `closeTunnel()` function.
 
 For more information, read this [blog post](https://blog.apify.com/tunneling-arbitrary-protocols-over-http-proxy-with-static-ip-address-b3a2222191ff).
 
@@ -579,7 +584,7 @@ Example:
 
 ```javascript
 const host = await createTunnel('http://bob:pass123@proxy.example.com:8000', 'service.example.com:356');
-// Prints something like "localhost:56836"
+// Prints something like "127.0.0.1:56836"
 console.log(host);
 ```
 
--- a/examples/apify_proxy_tunnel.js
+++ b/examples/apify_proxy_tunnel.js
@@ -15,7 +15,7 @@ import { closeTunnel, createTunnel, redactUrl } from 'proxy-chain';
     const TARGET_HOST = 'www.example.com:443';
 
     // Create tunnel for the service, this call will start local tunnel and
-    // return a string in format localhost:<selected-port>.
+    // return a string in format 127.0.0.1:<selected-port>.
     // Here we set "port" to 9999, but you can use 0 to get a random port.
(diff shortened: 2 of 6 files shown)


jancurn (approved):

examples/apify_proxy_tunnel.js
```diff
@@ -15,7 +15,7 @@ import { closeTunnel, createTunnel, redactUrl } from 'proxy-chain';
     const TARGET_HOST = 'www.example.com:443';
 
     // Create tunnel for the service, this call will start local tunnel and
-    // return a string in format localhost:<selected-port>.
+    // return a string in format 127.0.0.1:<selected-port>.
```
bliuchak, line 18: Will it work with IPv6?

src/anonymize_proxy.ts
```diff
@@ -30,13 +31,9 @@ export const anonymizeProxy = async (
         proxyUrl = options;
     } else {
         proxyUrl = options.url;
-        port = options.port;
+        port = options.port ?? 0;
```
bliuchak, line 35: Maybe it would make sense to add here a comment that explains port `0` behavior.

> When you pass 0 to `bind()` (or `server.listen(0)` in Node), you're telling the OS: "pick any free ephemeral port for me." The kernel assigns one from its ephemeral range, and you read back the real port afterward.

src/tcp_tunnel_tools.ts
```diff
@@ -2,9 +2,22 @@ import net from 'node:net';
 import { URL } from 'node:url';
 
 import { chain } from './chain.js';
+import { validateListenPort } from './utils/validate_listen_port.js';
 
 const runningServers: Record<string, { server: net.Server, connections: Set<net.Socket> }> = {};
 
+// The tunnel does not authenticate its clients and forwards the proxyUrl credentials
+// upstream, so it must not be reachable off the local machine by default.
+const DEFAULT_LISTEN_HOSTNAME = '127.0.0.1';
```
bliuchak, line 11: What about ipv6? Does it make sense to support both of them here?

src/tcp_tunnel_tools.ts
```diff
@@ -38,7 +53,20 @@ export async function createTunnel(
...
+    const listenPort = options?.port ?? 0;
+    const listenHostname = options?.hostname || DEFAULT_LISTEN_HOSTNAME;
+
+    validateListenPort(listenPort);
+
+    if (!isLoopbackHostname(listenHostname)) {
+        process.emitWarning(
+            `The tunnel is listening on "${listenHostname}", so it may be reachable from other machines.`
+            + ' It does not authenticate its clients and forwards the proxyUrl credentials upstream.',
+            'ProxyChainSecurityWarning',
```
bliuchak, line 65: Is this some kind of convention we'd like to introduce and use in the future?

bliuchak (approved):
Good job!

bliuchak:
@jirimoravcik could you please take a look at it (just in case we're not missing something)? 🙏

README.md
```diff
@@ -560,26 +560,31 @@ specified by the `proxyUrl` parameter.
...
    which means a random port will be selected.
-- `hostname: String` - Local hostname to listen at. By default `localhost`.
+- `hostname: String` - Local hostname to listen at. By default `127.0.0.1`.
 - `ignoreProxyCertificate` - For HTTPS proxy, ignore certificate errors in proxy requests. Useful for proxy with self-signed certificate. By default `false`.
 - `verbose: Boolean` - If `true`, the functions logs a lot. By default `false`.
 
-The result of the function is a local endpoint in a form of `hostname:port`.
+The result of the function is a local endpoint in a form of `hostname:port`,
+where `hostname` is the address the tunnel actually bound to. IPv6 addresses are
+bracketed, e.g. `[::1]:56836`. Pass this exact string to `closeTunnel()`.
```
jirimoravcik: > Pass this exact string to `closeTunnel()`.

I do not understand this sentence well, can we rephrase?

src/utils/validate_listen_port.ts
```diff
@@ -0,0 +1,5 @@
+export const validateListenPort = (port: number): void => {
```
jirimoravcik, line 2: Can we add a docstring please? One sentence is enough

src/tcp_tunnel_tools.ts
```diff
@@ -2,9 +2,22 @@ import net from 'node:net';
...
+// upstream, so it must not be reachable off the local machine by default.
+const DEFAULT_LISTEN_HOSTNAME = '127.0.0.1';
+
+const LOOPBACK_HOSTNAMES = new Set(['localhost', '::1', '0:0:0:0:0:0:0:1']);
+
+const isLoopbackHostname = (hostname: string) => {
+    const normalized = hostname.toLowerCase();
+
+    return LOOPBACK_HOSTNAMES.has(normalized) || normalized.startsWith('127.');
+};
```
jirimoravcik, line 19: By the way, there's a package for localhost IP, maybe you can get inspired https://github.com/tinovyatkin/is-localhost-ip/blob/master/index.js
Not sure if the current logic covers everything

jirimoravcik (approved):

src/tcp_tunnel_tools.ts
VojtaM39: This is just a default, you can still use ipv6. You can pass ::1 for hostname

src/utils/validate_listen_port.ts
VojtaM39: ff5d69825c31276dd9610301decbdbf894ab3416

src/anonymize_proxy.ts
VojtaM39: 509417b120db84c96274dcd7d869c699db853758

README.md
VojtaM39: 235ee06 - i have removed this sentence, as it was an unnecessary addition

examples/apify_proxy_tunnel.js
VojtaM39: Yes, you can optionally pass ipv6, this is just the default.

src/tcp_tunnel_tools.ts
VojtaM39: emitWarning should be the standard way to handle this in libraries, as it can be intercepted by the consumer

src/tcp_tunnel_tools.ts
VojtaM39: I think the package is too broad for this use-case and would mark LAN-reachable ranges as safe.

PR #682, opened 2026-08-28
chore: migrate tests from JS to TS

### LINKED ISSUE #674: Migrate tests from JS to TS
We should do this after migration from mocha to vitest: https://github.com/apify/proxy-chain/issues/648

VojtaM39:
Closes #674

- migrates all tests from JS to TS
- adds typecheck script and adds typecheck to CI
- removed tests from lint ignoring
- new test helpers


.github/workflows/check.yaml
```diff
@@ -43,6 +43,9 @@ jobs:
             - name: Lint code
               run: pnpm run lint
 
+            - name: Typecheck
+              run: pnpm run typecheck
```
daniil-poletaev: Our other repos have command named as `type-check`. Maybe we should be consistent here?

daniil-poletaev (approved):
Looks great!

.github/workflows/check.yaml
VojtaM39: 48b036d

bliuchak (approved):
Nice

PR #695, opened 2026-08-31
fix: remove outdated proxy package workaround

VojtaM39:
Removes unnecessary proxy package workaround that creates test flakiness: https://github.com/apify/proxy-chain/actions/runs/33449396551/job/99675639823?pr=694


bliuchak (approved):
Good catch!

PR #683, opened 2026-08-31
feat: add AGENTS.md

VojtaM39:
This PR adds basic AGENTS.md with underlying .md docs.
Fixes mistake in README.md, .nvmrc does not exist in this repo.

PR #701, opened 2026-09-07
chore: add automatic changelogs and release management

### LINKED ISSUE #705: Add changelog generation and release management

VojtaM39:
This PR modifies the release process so we don't have to do a manual bump and manually write changelogs.

The release happens by manually dispatching action with either specified bump type or auto bump, the action then creates release commit under service account, publishes to npm and prepends changelog.

To see how the changelog will look, check the dry run action.

Closes: #705


.github/workflows/release.yaml
```diff
@@ -1,32 +1,151 @@
-name: Check & Release
+# Everything that runs `pnpm publish` lives in this file on purpose. npm's trusted publisher
+# configuration for `proxy-chain` matches on the workflow filename, so moving a publish step
+# into another workflow breaks the OIDC handshake until that configuration is updated too.
+
+name: Release
 
 on:
-    # Push to master will deploy a beta version
+    # Every push to master publishes a `beta` prerelease.
```
bliuchak:
```suggestion
    # Every push to master publishes a `beta` pre-release.
```

VojtaM39:
@B4nan 
Could you also please check? We wanted to add a simple way to generate changelogs and simplify releases. This approach is heavily inspired from https://github.com/apify/apify-sdk-js/blob/master/.github/workflows/release.yaml.

bliuchak (approved):

daniil-poletaev (approved):

.github/workflows/release-dry-run.yaml
```diff
@@ -0,0 +1,86 @@
+# Shows what the next stable release would look like without writing anything, and checks that
+# generating the changelog still only prepends to the existing one. The header in CHANGELOG.md is
+# load-bearing - git-cliff strips exactly that string before prepending, so an edit to it silently
```
B4nan: > load-bearing

hey claude, how are you? :D

.github/workflows/release.yaml
```diff
@@ -1,32 +1,151 @@
...
+    release_commit:
+        name: Commit version and changelog
+        needs: [lint_and_test, release_metadata]
+        if: ${{ github.event_name == 'workflow_dispatch' }}
         runs-on: ubuntu-24.04
+        outputs:
+            commitish: ${{ steps.commit.outputs.commit_long_sha }}
 
         steps:
+            - name: Refuse to release from another branch
```
B4nan: This guard runs after `lint_and_test`, so a dispatch from a feature branch runs the whole test matrix before failing. Moving it into `release_metadata` (gated on `workflow_dispatch`) fails in seconds.

.github/workflows/release.yaml
```diff
@@ -1,32 +1,151 @@
...
+
+            # Commits through the GitHub API, so the commit ends up signed. The `[skip ci]` marker
+            # stops the beta publish from firing on the release commit.
+            - name: Create the release commit
+              id: commit
+              uses: apify/actions/signed-commit@v1.5.0
+              with:
+                  message: 'chore(release): v${{ needs.release_metadata.outputs.version_number }} [skip ci]'
+                  add: package.json CHANGELOG.md
+                  pull: '--rebase --autostash'
```
B4nan, line 133: Not blocking: a master push between dispatch and this step gets rebased under the release commit, so `publish` ships code that neither `lint_and_test` nor the changelog saw. Same trade-off as apify-client-js, fine if that is the accepted pattern.

B4nan (approved):
Lookg good, few nits claude found follow. 

Outside the diff: `docs/agents/git-and-prs.md` (lines 23-24) still says `latest` is published when a GitHub Release is published, and tells authors to bump `package.json` because `before-beta-release.cjs` fails otherwise. Both go away with this PR, and every agent-authored PR reads that file. It should point to `RELEASING.md` instead.

Also: the first `auto` dispatch computes `3.1.0`, and the only commit driving the minor is #683 (`feat: Add AGENTS.md`). Nothing user-facing changed there, so it should have been `chore` or `docs`. Either dispatch `patch` the first time or accept 3.1.0.

VojtaM39:
@bliuchak do we still want to release as 3.1.0?

Since https://github.com/apify/proxy-chain/pull/683#issuecomment-5573655578 should be chore, according to semver we should release 3.0.1.

.github/workflows/release.yaml
VojtaM39: 91094d3

B4nan:
Just pick the patch bump manually. Tbh, I am using explicit bumps all the time, never got used to the full automation. Some commits should get to the changelog, they are not bugs, yet they don't deserve a feature bump either. I like to stay in control.

bliuchak:
@VojtaM39 Well, I don't know the details of how this new "release" feature works right now. But I was expecting these changes (see list below) to be more minor than a patch.

List of things:
- Migrate tests from JS to Typescript
- Replace mocha with Vitest

I admit they're not adding any new functionality directly to proxy-chain, but I don't consider them small changes. We're replacing the entire test library and rewriting the tests in a different language.

Does it make sense?

P.S. I don't insist on a minor release. If we believe that the patch is better, let's do a patch. I don't have a strong opinion here.

B4nan:
> Migrate tests from JS to Typescript
> Replace mocha with Vitest

If you ask me, changelogs are for users, not maintainers, and none of this is user facing, to me those should not be part of the changelog. Both should have `test` commit type.

VojtaM39:
I also don't have a strong opinion, I think minor is fine. Just wanted to check 👍

B4nan:
In general, people are more likely to update to a patch bump, since all it should consist of is bug fixes, and who doesn't like bug fixes? But feature releases are a different story, minor bumps might see slower adoption (and also higher expectations, people might be sad seeing a minor that doesn't have any user facing changes 🙃).

But yes, it's surely not that important, do as you wish :]

